PR #2, opened 2015-05-21
Cleanup / Review

chirimoya:


lib/Event/ConfigureOptionsEvent.php
```diff
@@ -15,13 +15,22 @@
...
 {
+    /**
+     * @var OptionsResolverInterface
+     */
     private $options;
 
+    /**
+     * @param OptionsResolverInterface $options
+     */
     public function __construct(OptionsResolverInterface $options)
```
chirimoya: @dantleech ... `OptionsResolverInterface` is deprecated since Symfony 2.6 and will be removed 3.0. Not sure why, but should we not use the `OptionsResolver` as suggested?

dantleech: Sure

lib/Query/Query.php
```diff
@@ -30,14 +30,42 @@ class Query
...
+
+    /**
+     * @var int
+     */
     private $maxResults;
+
+    /**
+     * @var int
+     */
     private $firstResult;
```
chirimoya, line 61: Why `$firstResult` and not `$offset`?

lib/Query/Query.php
```diff
@@ -30,14 +30,42 @@ class Query
     const HYDRATE_DOCUMENT = 'document';
     const HYDRATE_PHPCR = 'phpcr_node';
 
-    private $hydrationMode;
```
chirimoya, line 33: Removed `$hydrationMode` property cause it never was used.

lib/Query/Query.php
dantleech, line 61: Its inherited from the PHPCR-ODM (this actually started off as a copy of it).

lib/Query/Query.php
dantleech, line 33: ok. It might be added back in when I add the qurey builder.

chirimoya:
@dantleech ... method `dereference(NodeInterface $node)` from `RemoveSubscriber` never is used. Can be removed? ... https://github.com/sulu-io/sulu-document-manager/pull/2/files#diff-5b6baea4af453272f4861d25be3b85d5L94

chirimoya:
@dantleech 

```
foreach (array_keys($parentNode->getNodes()) as $name) {
    ...
```

array_keys expects an array, but `getNodes()` returns an `Iterator`. Does this work as expected?

 ... https://github.com/sulu-io/sulu-document-manager/pull/2/files#diff-e70eded7ee1f3b0becdb1eefab8ef28aL116

chirimoya:
Also should the result of array_keys be "cached" in a variable?

dantleech:
>  method dereference(NodeInterface $node) from RemoveSubscriber never is used.

yes

> array_keys expects an array, but getNodes() returns an Iterator. Does this work as expected?

I assume so. this code was copied verbatim from the ContentMapper.

> Also should the result of array_keys be "cached" in a variable?

I don't think so, we could add some static caching, but this should already be cached by jackalope.

chirimoya:
> >  method dereference(NodeInterface $node) from RemoveSubscriber never is used.
> 
>  yes

yes, remove? ;-)

chirimoya:
@dantleech ... travis tests are failing cause of issues with `phpbench`. can you have a look? thx

lib/Query/Query.php
chirimoya, line 61: Should we rename or not?

dantleech, line 61: I would say leave it as it is. It is also named this way in Doctrine ORM, so changing it would just be more cognitive overhead.

dantleech:
@chirimoya yeah. will try and make a tag of PHPBench soon.

dantleech:
and yeah, we can remove the `dereference` method. Originally I was going to have the DocumentManager handle dereferencing nodes (i.e. explicitly removing any existing references to a node prior to delete), but our use case was more specialized.

chirimoya:
@dantleech done. can you review.

chirimoya:
@dantleech ... good job so far! looking forward to evolve the document manager.

dantleech:
Fixed the benchmarks.

lib/ClassNameInflector.php
```diff
@@ -15,12 +15,15 @@
 
 /**
  * This is a hack to statically use the ClassNameInflector to
- * retireve the "real" classnames for proxy objects.
+ * retrieve the "real" class-names for proxy objects.
```
dantleech: should be `class names`, but its fine.

lib/Query/Query.php
chirimoya, line 61: ack

lib/ClassNameInflector.php
chirimoya: fixed

lib/DocumentRegistry.php
```diff
@@ -12,27 +22,32 @@ class DocumentRegistry
...
      * @var array
      */
-    private $documentNodeMap;
+    private $nodeMap = array();
 
     /**
      * @var array
      */
-    private $nodeDocumentMap;
+    private $nodeDocumentMap = array();
```
dantleech, line 40: Hmm. good catch.

dantleech:
+1

chirimoya:
@dantleech are you going to merge this PR

PR #3, opened 2015-05-27
Query builder

dantleech:
Initial work for supporting a query builder.

This PR reuses the PHPCR-ODM query builder, and depends on a pull request to make it extensible.

We will need to implement a custom QueryBuilder class in Sulu in order to be able to filter based on structure types.

Example:

``` php
$qb = $documentManager->createQueryBuilder();
$qb->from()->document('page#overview', 'o');
$qb->where()->field('o.#title')->literal('Hello');
$results = $qb->getQuery()->execute();
```

### DIFF
--- a/tests/Functional/DocumentManager/QueryBuilderTest.php
+++ b/tests/Functional/DocumentManager/QueryBuilderTest.php
@@ -1,0 +1,62 @@
+<?php
+
+/*
+ * This file is part of the Sulu CMS.
+ *
+ * (c) MASSIVE ART WebServices GmbH
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace Sulu\Component\DocumentManager\Tests\Functional\DocumentManager;
+
+use Sulu\Component\DocumentManager\Tests\Functional\BaseTestCase;
+
+class DocumentManagerTest extends BaseTestCase
+{
+    public function setUp()
+    {
+        $this->initPhpcr();
+    }
+
+    private function createDocument($title)
+    {
+        $manager = $this->getDocumentManager();
+        $document = $manager->create('full');
+
+        $manager->persist($document, 'en', array(
+            'path' => self::BASE_PATH . '/' . $title,
+            'auto_create' => true,
+        ));
+        $manager->flush();
+
+        return $document;
+    }
+
+    /**
+     * Test select from alias.
+     */
+    public function testFromAlias()
+    {
+        $this->createDocument('Hello');
+        $manager = $this->getDocumentManager();
+        $builder = $manager->createQueryBuilder();
+        $query = $builder->from()->document('full', 'p')->end()->getQuery();
+        $results = $query->execute();
+        $this->assertCount(1, $results);
+    }
+
+    /**
+     * Test select from document class.
+     */
+    public function testFromDocumentClass()
+    {
+        $this->createDocument('bar');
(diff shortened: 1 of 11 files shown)


lib/Query/BuilderConverterSulu.php
```diff
@@ -0,0 +1,225 @@
...
+        }
+
+        if ($maxResults = $builder->getMaxResults()) {
+            $query->setMaxResults($maxResults);
+        }
+
+        return $query;
+    }
+
+    public function walkSelect(AbstractNode $node)
```
dantleech: This method cna be removed.

lib/Query/BuilderConverterSulu.php
```diff
@@ -0,0 +1,225 @@
...
+use Doctrine\ODM\PHPCR\Query\Builder\BuilderConverterPhpcr;
+use Doctrine\DBAL\Query\QueryBuilder;
+use Doctrine\ODM\PHPCR\Query\Builder\SourceDocument;
+use Doctrine\ODM\PHPCR\Query\Builder\OperandDynamicField;
+use Sulu\Component\DocumentManager\MetadataFactoryInterface;
+
+/**
+ * Class which converts a Builder tree to a PHPCR Query
+ */
+class BuilderConverterSulu extends BuilderConverterPhpcr
```
dantleech: Should now extend the BuilderConverterBase class when (if) the relevanyt PHPCR-ODM PR is merged

tests/Functional/DocumentManager/QueryBuilderTest.php
```diff
@@ -0,0 +1,81 @@
...
+        $builder = $manager->createQueryBuilder();
+        $query = $builder->from()->document('Sulu\Component\DocumentManager\Tests\Functional\Model\FullDocument', 'p')->end()->getQuery();
+        $results = $query->execute();
+        $this->assertCount(1, $results);
+    }
+
+    /**
+     * Search for document by title
+     */
+    public function testSearchByTitle()
```
dantleech: Can't actually do this at the moment as there is no way in this component to map any values to the PHPCR node.

lib/Query/BuilderConverterSulu.php
```diff
@@ -0,0 +1,244 @@
...
+     * @var EventDispatcherInterface
+     */
+    private $eventDispatcher;
+
+    /**
+     * @var Metadata[]
+     */
+    protected $documentMetadata = array();
+
+    public function __construct(
```
dantleech: +docs

lib/Query/BuilderConverterSulu.php
```diff
@@ -0,0 +1,244 @@
...
+        }
+
+        if ($maxResults = $builder->getMaxResults()) {
+            $query->setMaxResults($maxResults);
+        }
+
+        return $query;
+    }
+
+    protected function applySourceConstraints(QueryBuilder $builder)
```
dantleech: protected after public

dantleech:
Th PHPCR-ODM PR this depends on has now been merged, but it couples us to the next major version (`1.3`)

dantleech:
TODO:
- Subscribers need to communicate fields that they map in order that they can be queries on, for example `changer`, `updated`, `published`, etc.
- Need to select from "abstract" types, e.g. both `page` and `home` document types.

dantleech:
I think we need to implement Metadata field mapping before the query builder can be effectively implemented.

dantleech:
Merged into develop.

PR #5, opened 2015-06-02
Added path builder

dantleech:
This PR introduces the path builder.

The path builder allows you to construct paths by providing an array of path segments. If a path segment is demarcated with `%` then it will be resolved by the `PathSegmentRegistry`.

``` php
$builder->build('%base%', 'sulu_io', '%content%', 'path/to/article');

echo $path; // /cmf/sulu_io/contents/path/to/article
```

This replaces the functionality of the Sulu `SessionManager`.

Note that I had originally planned to make something cleverer, e.g. `$builder->segment('webspace', $webspace->getKey())->add('foobar')`. This enforces a kind of "schema" but is inherently more complicated.

I think the approach I have done here is flexible and simple, and does not require any additional classes or configuration - at the cost of being less schematic.

### DIFF
--- a/lib/Collection/ReferrerCollection.php
+++ b/lib/Collection/ReferrerCollection.php
@@ -79,7 +79,7 @@ class ReferrerCollection extends AbstractLazyCollection
         //       initialized, but if we don't do this, we won't know how many items are in the
         //       collection, as one node could have many referring properties.
         foreach ($references as $reference) {
-            /** @var PropertyInterface $reference */
+            /* @var PropertyInterface $reference */
             $referrerNode = $reference->getParent();
             $this->documents[$referrerNode->getIdentifier()] = $referrerNode;
         }
--- a/lib/PathBuilder.php
+++ b/lib/PathBuilder.php
@@ -1,0 +1,78 @@
+<?php
+
+namespace Sulu\Component\DocumentManager;
+
+/**
+ * The path builder provides a way to create paths from templates
+ */
+class PathBuilder
+{
+    /**
+     * @var PathSegmentRegistry
+     */
+    private $registry;
+
+    /**
+     * @param PathSegmentRegistry $registry
+     */
+    public function __construct(PathSegmentRegistry $registry)
+    {
+        $this->registry = $registry;
+    }
+
+    /**
+     * Build a path from an array of path segments.
+     *
+     * Segments demarcated by "%" characters will be interpreted as path
+     * segment *names* and their value will be resolved from the PathSegmentRegistry.
+     *
+     * Other segments will be interpreted literally.
+     *
+     * The following:
+     *
+     * ````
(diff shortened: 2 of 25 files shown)


tests/Bench/PathBuilderBench.php
```diff
@@ -0,0 +1,44 @@
...
+        return array(
+            array(
+                'elements' => array('one', 'two', 'three')
+            ),
+            array(
+                'elements' => array('%one', '%two%', 'three'),
+            ),
+        );
+    }
+
```
dantleech: Extra line

tests/Unit/PathBuilderTest.php
```diff
@@ -0,0 +1,65 @@
...
+    }
+
+    /**
+     * It should replace "//" with "/"
+     */
+    public function testBuildNoDoubleSlash()
+    {
+        $this->assertEquals('/hello/world', $this->pathBuilder->build(array('hello', '', '', 'world')));
+    }
+}
```
dantleech, line 63: Add test for nested paths, e.g. `array('foo', 'foo/bar/bar')`

tests/Unit/PathBuilderTest.php
```diff
@@ -0,0 +1,65 @@
...
+        ));
+        $this->pathBuilder = new PathBuilder($pathRegistry);
+    }
+
+    /**
+     * @description Build path 1000 times
+     * @revs 1000
+     * @beforeMethod setUp
+     */
+    public function benchBuild()
```
dantleech: Remove this

lib/PathBuilder.php
```diff
@@ -0,0 +1,78 @@
...
+     *
+     * Will yield `/cms/hello/articleDirectory` where `%base%` is "cms" and
+     * `%articles` is "articleDirectory"
+     *
+     * @see Sulu\Component\DocumentManager\PathSegmentRegistry
+     *
+     * @param array $segments
+     * @return string
+     */
+    public function build($segments)
```
dantleech: Add type hint `array`

lib/PathBuilder.php
```diff
@@ -0,0 +1,78 @@
+<?php
+
```
dantleech, line 2: Run the CS fixer.

lib/Collection/ReferrerCollection.php
```diff
@@ -79,7 +79,7 @@ protected function initialize()
         //       initialized, but if we don't do this, we won't know how many items are in the
         //       collection, as one node could have many referring properties.
         foreach ($references as $reference) {
-            /** @var PropertyInterface $reference */
+            /* @var PropertyInterface $reference */
```
danrot, line 82: This change is wrong, the comment is not recognized as PHPDoc anymore.

PR #6, opened 2015-06-03
introduced DocumentManagerInterface

danrot:
This PR introduces an interface for the `DocumentManager`, so that the implementation can be easily exchanged later (e.g. for benchmarking).


dantleech:
We will probably need to update a couple of classes in `sulu/sulu#refactoring` also.

PR #15, opened 2015-06-19
added options to all required events

danrot:
I have introduced an `EventOptionsTrait`, which I have used instead of the own implementation spread over multiple events, and introduced in more events I needed.

**tasks:**
- [ ] test coverage

**informations:**

| q | a |
| --- | --- |
| Fixed tickets | none |
| BC breaks | none |
| Documentation PR | none |

### DIFF
--- a/lib/Event/EventOptionsTrait.php
+++ b/lib/Event/EventOptionsTrait.php
@@ -1,0 +1,47 @@
+<?php
+/*
+ * This file is part of Sulu
+ *
+ * (c) MASSIVE ART WebServices GmbH
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+namespace Sulu\Component\DocumentManager\Event;
+
+/**
+ * This trait adds options to an event
+ */
+trait EventOptionsTrait
+{
+    /**
+     * This array is used as key value storage for the options
+     * @var array
+     */
+    protected $options = array();
+
+    /**
+     * Returns all the options for the event
+     *
+     * @return array
+     */
+    public function getOptions()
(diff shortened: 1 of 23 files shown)


lib/Event/EventOptionsTrait.php
```diff
@@ -0,0 +1,47 @@
...
+    }
+
+    /**
+     * Returns the option with the given name
+     * @param string $name The name of the option
+     * @param mixed $default The return value in case the option is not set
+     *
+     * @return mixed The value of the option
+     */
+    public function getOption($name, $default = null) {
```
dantleech: `{` on new line

lib/Event/QueryCreateEvent.php
```diff
@@ -39,12 +41,14 @@ class QueryCreateEvent extends AbstractEvent
     /**
      * @param string $innerQuery
      * @param string $locale
+     * @param array $options
      * @param null|string $primarySelector
      */
-    public function __construct($innerQuery, $locale, $primarySelector = null)
+    public function __construct($innerQuery, $locale, array $options, $primarySelector = null)
```
dantleech: Should `$options` be required here?

dantleech:
Have we got sufficient test coverage to ensure that the options are being passed to the HydrateEvent from the QueryCreateEvent?

danrot:
I don't think so... Would it be ok for you to just create some random configuration which is not used at all? Just to test if it is passed?

lib/Event/QueryCreateEvent.php
danrot: I guess not :)

PR #17, opened 2015-06-22
Allow persisting without locale

dantleech:
Allow persisting without locale.

Note that this will need to be cherry picked into master.


dantleech:
Cherry picked into develop

PR #28, opened 2015-07-20
extended reorder event with node

danrot:
**tasks:**
- [x] test coverage
- [x] gather feedback for my changes
- [x] submit changes to the documentation

**informations:**

| q | a |
| --- | --- |
| Fixed tickets | none |
| BC breaks | none |
| Documentation PR | none |

### DIFF
--- a/lib/Event/ReorderEvent.php
+++ b/lib/Event/ReorderEvent.php
@@ -11,7 +11,9 @@
 
 namespace Sulu\Component\DocumentManager\Event;
 
-class ReorderEvent extends AbstractDocumentEvent
+use PHPCR\NodeInterface;
+
+class ReorderEvent extends AbstractMappingEvent
 {
     /**
      * @var string
@@ -30,7 +32,7 @@ class ReorderEvent extends AbstractDocumentEvent
      */
     public function __construct($document, $destId, $after)
     {
-        parent::__construct($document);
+        $this->document = $document;
         $this->destId = $destId;
         $this->after = $after;
     }
@@ -62,5 +64,13 @@ class ReorderEvent extends AbstractDocumentEvent
     public function getAfter()
     {
         return $this->after;
+    }
+
+    /**
+     * @param NodeInterface $node
+     */
+    public function setNode(NodeInterface $node)
+    {
+        $this->node = $node;
     }
 }
--- a/lib/Subscriber/Core/RegistratorSubscriber.php
+++ b/lib/Subscriber/Core/RegistratorSubscriber.php
@@ -17,6 +17,7 @@ use Sulu\Component\DocumentManager\Event\ClearEvent;
 use Sulu\Component\DocumentManager\Event\HydrateEvent;
 use Sulu\Component\DocumentManager\Event\PersistEvent;
 use Sulu\Component\DocumentManager\Event\RemoveEvent;
+use Sulu\Component\DocumentManager\Event\ReorderEvent;
 use Sulu\Component\DocumentManager\Events;
 use Symfony\Component\EventDispatcher\EventSubscriberInterface;
 
@@ -60,6 +61,7 @@ class RegistratorSubscriber implements EventSubscriberInterface
             ),
             Events::REMOVE => array('handleRemove', 490),
             Events::CLEAR => array('handleClear', 500),
+            Events::REORDER => array('handleNodeFromRegistry', 510),
         );
     }
 
@@ -150,9 +152,9 @@ class RegistratorSubscriber implements EventSubscriberInterface
     /**
      * If the node for the persisted document is in the registry.
      *
-     * @param PersistEvent $event
+     * @param PersistEvent|ReorderEvent $event
      */
-    public function handleNodeFromRegistry(PersistEvent $event)
+    public function handleNodeFromRegistry($event)
     {
         if ($event->hasNode()) {
             return;
--- a/tests/Unit/Subscriber/Core/RegistratorSubscriberTest.php
+++ b/tests/Unit/Subscriber/Core/RegistratorSubscriberTest.php
@@ -7,10 +7,46 @@ use Sulu\Component\DocumentManager\DocumentRegistry;
 use Sulu\Component\DocumentManager\Event\HydrateEvent;
 use Sulu\Component\DocumentManager\Event\PersistEvent;
 use Sulu\Component\DocumentManager\Event\RemoveEvent;
+use Sulu\Component\DocumentManager\Event\ReorderEvent;
 use Sulu\Component\DocumentManager\Subscriber\Core\RegistratorSubscriber;
 
 class RegistratorSubscriberTest extends \PHPUnit_Framework_TestCase
 {
+    /**
+     * @var DocumentRegistry
+     */
+    private $registry;
+
+    /**
+     * @var RegistratorSubscriber
+     */
+    private $subscriber;
+
+    /**
+     * @var NodeInterface
+     */
+    private $node;
+
+    /**
+     * @var \stdClass
+     */
+    private $document;
+
+    /**
+     * @var HydrateEvent
+     */
+    private $hydrateEvent;
+
+    /**
+     * @var PersistEvent
+     */
+    private $persistEvent;
+
+    /**
+     * @var RemoveEvent
+     */
+    private $removeEvent;
+
     public function setUp()
     {
         $this->registry = $this->prophesize(DocumentRegistry::class);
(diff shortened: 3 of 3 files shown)


lib/Subscriber/Core/RegistratorSubscriber.php
```diff
@@ -230,4 +211,25 @@ private function handleRegister(AbstractMappingEvent $event)
 
         $this->documentRegistry->registerDocument($document, $node, $locale);
     }
+
+    /**
+     * If the node for the persisted document is in the registry.
+     *
+     * @param PersistEvent|ReorderEvent $event
+     */
+    public function handleNodeFromRegistry($event)
```
dantleech: Why move this? `private` functions are intentionally placed below `public` functions.

lib/Event/ReorderEvent.php
```diff
@@ -30,7 +32,7 @@ class ReorderEvent extends AbstractDocumentEvent
      */
     public function __construct($document, $destId, $after)
     {
-        parent::__construct($document);
+        $this->document = $document;
```
dantleech, line 35: Why change this? We should call the parent constructor, this class file does not know about the document (`$document`  could actually probably be `private`).

tests/Unit/Subscriber/Core/RegistratorSubscriberTest.php
```diff
@@ -104,6 +140,17 @@ public function testNodeFromRegistry()
         $this->subscriber->handleNodeFromRegistry($this->persistEvent->reveal());
     }
 
+    public function testReorderNodeFomRegistry()
```
dantleech: Add comment as with other tests.

dantleech: i.e. `test reorder node from registry` -- what does that mean? You are testing that `The node should be available from the event` or?

lib/Event/ReorderEvent.php
danrot, line 35: Because I also changed the class it extends from, and the new one doesn't have a constructor. Don't like this inheritance structure at all, would prefer to replace that with interfaces and traits, but this seemed to be the cleanest solution for now.

lib/Subscriber/Core/RegistratorSubscriber.php
danrot: Sry, have tested something different first, will move it back.

tests/Unit/Subscriber/Core/RegistratorSubscriberTest.php
danrot: As already said I am against adding these comments, most of the time they are just duplicating information. The name already shows that it is testing if the node is set from the registry on the reorder event. If you think the name is not sufficient I would still prefer to change the name, instead of adding more comments.

lib/Event/ReorderEvent.php
dantleech, line 35: No, it was not something that was "designed", it just evolved like this
and, no, it isn't very satisfying.  But I don't think it is important to
change this. Introducing traits and interfaces would be over-the-top in
this situation. Events are not polymorphic (i.e.  interfaces not
required). Traits might be useful but I don't think its worth the
effort.

tests/Unit/Subscriber/Core/RegistratorSubscriberTest.php
dantleech: The human language description has the potential to be much more
preceise and expressive than a method name. This is how all the tests in
this library have been written.

lib/Event/ReorderEvent.php
danrot, line 35: Actually you are using them polymorphic, since you have type hinted the `handleHydrate` method parameter with `AbstractMappingEvent` sometimes, so that you can pass a `PersistEvent` in there. Therefore it would totally make sense to me, to introduce these traits and interfaces, and I guess I will do that in one of the next iterations. Would make some code reuse between these elements also easier, e.g. there is no need to implement `setNode` multiple times.

tests/Unit/Subscriber/Core/RegistratorSubscriberTest.php
danrot: @chirimoya What's your opinion on that?

danrot: I am just afraid that these comments won't match the method name any time in the future, and people will get confused by this. Already happened in actual code to me from time to time, and I could imagine that people are spending even less focus on this in the tests.

chirimoya: Currently there is no guideline to add comments for tests. IMHO If an additional comment adds value, it definitely makes sense. If it's the same like the method name, not really.

lib/Event/ReorderEvent.php
dantleech, line 35: To me it makes no sense to have interfaces for events. When I say
"polymorphic" I mean they are not intended to be replaced by the user by
a different implementation nor should this ever happen.

If there were a requirement to change this architecture lets think about
that when we meet it. I can see no reason to refactor this just for the
sake of refactoring to a solution which may or not be better than the
current solution.

tests/Unit/Subscriber/Core/RegistratorSubscriberTest.php
dantleech: In anycase, this is how the current code has been written. I did not
write it like this for fun. This approach will ensure that people can
understand the tests and the purpose of each test. Please add the
comment.

lib/Event/ReorderEvent.php
danrot, line 35: The point is that it already would be better than the current situation. My proposed solution would remove the `setNode` redundancy in the `PersistEvent` and `ReorderEvent`, and would give me the possibility to have a matching type hint in the `handleNodeFromRegistry` function of the `RegistratorSubscriber` (the other possibility would be to have two separate functions). It would also allow to easily use the `RegistratorSubscriber` with other events, which might come in the future.

And just because there is no intention to let user add or change these event (without saying this would be a bad idea, haven't thought about it too much), it still doesn't mean it is not polymorphic. As I already said this would make sense in the `handleNodeFromRegistry` function to have a `SetNodeInterface` (would have to come up with a better name though), and you've been already using it on your own with the `AbstractMappingEvent` type hint in some `handleHydrate` methods. There it would also be better to have a cleaner separation.

tests/Unit/Subscriber/Core/RegistratorSubscriberTest.php
danrot: Actually I just wanted to give up on this, but now I have seen the following on another PR:
image

This is exactly what happens if we force this policy, and this is also the reason why I have a very strong opinion against it.

lib/Event/ReorderEvent.php
dantleech, line 35: There will only ever be a single implementation of any given event class. Introducing interfaces for the sake of the type hint that is already implemented is solving a problem that does not exist. If there is a valid case then lets consider this, but currently I do not see the problem. I would rather the code be not-perfect than it be prematurely over enginered,

tests/Unit/Subscriber/Core/RegistratorSubscriberTest.php
dantleech: That is not the type of comment that is given here. These comments are specifications:

```
/**
 * It should retrieve a media entity by ID
 */
```

Would be valid. This is a _policy_ which helps write better tests and is the reason I have adopted it here.

danrot: This comment also doesn't add any value. This is a policy which leads to comments like the ones above. I'll add it now, but I want this be discussed now, since you have introduced this "policy" without any discussion in the team. /cc @chirimoya @wachterjohannes 

As already said I have a VERY strong opinion against that.

lib/Event/ReorderEvent.php
danrot, line 35: You can see the use case already in this PR: https://github.com/sulu-io/sulu-document-manager/pull/28/files#diff-759b4f020ccaf54b3361e72dbe3f7dfbR155

This solution is quite simple, and I would not call it over engineered.

tests/Unit/Subscriber/Core/RegistratorSubscriberTest.php
dantleech: I introduced this policy when writing this library. It does add value,
otherwise I would not have added them. Please indulge me.

danrot: Looks like we won't agree on this one :smile:

PR #29, opened 2015-07-22
added styleci config file

danrot:


danrot:
What should we do about the `tests` and `Tests` namespace here? I guess we would have to disable the PSR-0 fixer in styleci: https://styleci.readme.io/docs/fixers

Is this one needed at all? If the namespaces are wrong the tests probably would fail anyway...

PR #30, opened 2015-08-12
removed the locale as mandatory parameter from persist

### LINKED ISSUE #16: Make locale non-mandatory for persist.
Currently it is mandatory to pass the locale to `persist`. But in reality it is not required as we always fallback to the default locale.

danrot:
This PR removes the locale as mandatory parameter, because sometimes you only want to persist changes that are not locale specific.

**tasks:**
- [ ] test coverage

**informations:**

| q | a |
| --- | --- |
| Fixed tickets | fixes #16 |
| BC breaks | none |
| Documentation PR | none |

### DIFF
--- a/lib/Event/EventOptionsTrait.php
+++ b/lib/Event/EventOptionsTrait.php
@@ -11,6 +11,8 @@
 
 namespace Sulu\Component\DocumentManager\Event;
 
+use Symfony\Component\OptionsResolver\OptionsResolver;
+
 /**
  * This trait adds options to an event.
  */
@@ -19,14 +21,14 @@ trait EventOptionsTrait
     /**
      * This array is used as key value storage for the options.
      *
-     * @var array
+     * @var OptionsResolver
      */
-    protected $options = [];
+    protected $options;
 
     /**
      * Returns all the options for the event.
      *
-     * @return array
+     * @return OptionsResolver
      */
     public function getOptions()
     {
--- a/lib/Subscriber/Behavior/Audit/BlameSubscriber.php
+++ b/lib/Subscriber/Behavior/Audit/BlameSubscriber.php
@@ -92,8 +92,13 @@ class BlameSubscriber implements EventSubscriberInterface
             return;
         }
 
-        $node = $event->getNode();
         $locale = $event->getLocale();
+
+        if (!$locale) {
+            return;
+        }
+
+        $node = $event->getNode();
 
         if (!$this->getCreator($node, $locale)) {
             $name = $this->encoder->localizedSystemName(self::CREATOR, $locale);
@@ -106,7 +111,7 @@ class BlameSubscriber implements EventSubscriberInterface
         $this->handleHydrate($event);
     }
 
-    private function getUserId(array $options)
+    private function getUserId($options)
     {
         if ($options['user']) {
             return $options['user'];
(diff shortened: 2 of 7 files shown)


lib/Event/EventOptionsTrait.php
```diff
@@ -19,14 +21,14 @@
     /**
      * This array is used as key value storage for the options.
      *
-     * @var array
+     * @var OptionsResolver
```
dantleech, line 24: This is not an options resolver, its an array iirc.

dantleech, line 24: The options resolver only "resolves" an array of options to according to a configuration. The options themselves are always an array.

danrot, line 24: It's not, on these options methods like `setDefault` are called.

lib/Subscriber/Behavior/Audit/BlameSubscriber.php
```diff
@@ -92,9 +92,14 @@ public function handlePersist(PersistEvent $event)
             return;
         }
 
-        $node = $event->getNode();
         $locale = $event->getLocale();
 
+        if (!$locale) {
+            return;
```
dantleech, line 98: /Should this ever happen? Should it not already have the default locale by this point?

lib/Subscriber/Behavior/Audit/TimestampSubscriber.php
```diff
@@ -59,14 +59,20 @@ public function handlePersist(PersistEvent $event)
             return;
         }
 
+        $locale = $event->getLocale();
```
dantleech, line 62: Same qiestion.

lib/Subscriber/Behavior/Audit/BlameSubscriber.php
danrot, line 98: I didn't do that in the first place, and then there was a properties called like `i18n:-changer`, so I would say this is necessary. And even if it would be it should not update any data from a language which was not really edited.

lib/Event/EventOptionsTrait.php
dantleech, line 24: Yeah, your right. In this case I would suggest to change the variable to "resolver" and remove the default value. "options" is just a misleading variable name.

danrot, line 24: I've removed the default value, but I am not sure if it is a good idea to change the name of the variable and method, since it feels strange in some places. E.g. if you call `getResolver` just to get the options in a subscriber, and the fact that there is also already a `getOption` method, which should not be renamed. Having two different method names here for more or less the same thing (only the count differs) feels quite strange...

dantleech, line 24: If I understand your point: you cannot determine the options from the OptionsResolver, it is there to be configured by the subscribers, it doesn't contain any option data, so you can't "get the options" from the resolver.

danrot, line 24: You can, exactly the same way it is done in this class' `getOption` method. The `OptionsResolver` implements `\ArrayAccess`.

lib/Subscriber/Behavior/Audit/BlameSubscriber.php
dantleech, line 98: ok

lib/Event/EventOptionsTrait.php
dantleech, line 24: ok I see. It kinda sucks, but OK lets keep the name.

dantleech, line 24: Now I think I remember having this dilema when I created the class

PR #34, opened 2015-09-14
added own LazyLoadingGhostFactory to make directory configureable in DI

danrot:


lib/Subscriber/Core/RegistratorSubscriber.php
```diff
@@ -218,6 +218,7 @@ public function handleClear(ClearEvent $event)
      *
      * @param AbstractMappingEvent $event
      */
+
```
wachterjohannes: remove this empty line

lib/LazyLoadingGhostFactory.php
```diff
@@ -0,0 +1,32 @@
...
+ */
+class LazyLoadingGhostFactory extends BaseLazyLoadingGhostFactory
+{
+    public function __construct($cacheDirectory)
+    {
+        if (!is_dir($cacheDirectory)) {
+            mkdir($cacheDirectory);
+        }
+
+        $configuration = new Configuration();
```
wachterjohannes, line 27: couldn`t we build this objectwith DI?

lib/LazyLoadingGhostFactory.php
```diff
@@ -0,0 +1,32 @@
...
+
+/**
+ * Extends the LazyLoadingGhostFactory of the ProxyManager, to make it configureable according to the needs of Sulu.
+ */
+class LazyLoadingGhostFactory extends BaseLazyLoadingGhostFactory
+{
+    public function __construct($cacheDirectory)
+    {
+        if (!is_dir($cacheDirectory)) {
+            mkdir($cacheDirectory);
```
dantleech: Note that there was a recent security vulnerability in Doctrine whereby cache files created with the permissions 777 could be modified by any user on the system to execute any code as the web user. We should probably review Sulu for this security problem generally.

lib/Subscriber/Core/RegistratorSubscriber.php
danrot: This line was introduced by styleci... Have to check why this happens...

lib/LazyLoadingGhostFactory.php
danrot: I have checked it, the permissions are fine: 
image

lib/LazyLoadingGhostFactory.php
```diff
@@ -0,0 +1,32 @@
...
+
+/**
+ * Extends the LazyLoadingGhostFactory of the ProxyManager, to make it configureable according to the needs of Sulu.
+ */
+class LazyLoadingGhostFactory extends BaseLazyLoadingGhostFactory
+{
+    public function __construct($cacheDirectory)
+    {
+        if (!is_dir($cacheDirectory)) {
+            mkdir($cacheDirectory, 0777, true);
```
dantleech, line 24: `777` is the issue here: http://www.doctrine-project.org/2015/08/31/security_misconfiguration_vulnerability_in_various_doctrine_projects.html but it should probably be fixed globally in another PR.

Also, why not remove this class and just configure the `Configuration` in the DI container and set the cache directory there? The cache directory would typically already exist (e.g. `app/cache`).

danrot:
Obsolete because of change suggested in https://github.com/sulu-io/sulu/pull/1576#discussion_r39388514

PR #37, opened 2015-09-21
[WIP] Improvement/filing subscriber

stemod:


dantleech:
Think you are doing this on the wrong branch or something, but I think what you propose here would be better if it were configurable. What is the use case?

PR #38, opened 2015-09-21
Added possibility to use AbstractFilingSubscriber for different document types

stemod:
Merging this branch requires merging branch feature/video-manager of  sulu/sulu.
https://github.com/sulu-io/sulu/pull/1611


dantleech:
Please add a description. What is the use case?

danrot:
@stemod The tests have also to be fixed in this repository. And can you link the PR in sulu/sulu here as well?

@dantleech Making it configureable was the first idea, but what you have built before doesn't work that way. The subscriber had the entire path, and every document having the `StructureFilingBehavior` ended up with this single path. So this is the solution we came up with. The use case is to have another type of document which are structured in a similar way as the snippets.

If I think about it, it doesn't make a lot of sense to me to have another configuration for  this, since the alias is already unique, and adding even more configuration makes stuff only a lot more complicated, without adding any value. What would probably be a good idea is to put these typed nodes like `cmf/snippets` into an own namespace.

dantleech:
Not sure I understand this. The `AliasFilingSubscriber` which uses this `AbstractFilingSubscriber` already uses the alias name as the parent of the document being saved. So the `AliasFilingSubscriber` would now save:

```
/snippets/hotel/hotels/the-grand-budapest
```

Or?

danrot:
Hm, have to check again, but if I remember correctly it was correct. Previously `snippets` was hardcoded in the DI configuration, it will be deleted there.

@stemod Can you check this, and also create the PR for sulu/sulu, so that this entire stuff will be more clear to @dantleech ?

dantleech:
As long as I understand:

`snippets` is hard coded in the DI config, because it is the base path.
When using the alias filing subscriber it will use the alias the next
element of the path.

However - I think I see what you want to do: You want:

```
/%base_path%/%document_alias%/%structure_type%
```

or:

```
/cmf/snippets/hotel
```

In this case you should modify the `StructureFilingSubscriber` in
`sulu/sulu` and
override the getParentName to return the document alias before the
structure type (it will also work if you return a path
segment such as `<document-alias>/<structure-type>`).

This subscriber works fine in the context of a standalone document
manager I think.

If you wanted to do something else, you just need to _extend_ the
`AbstractFilingSubscriber` and override the `getParentPath` method.

danrot:
No, that doesn't work, because then we would have to introduce a lot of different behaviors, which would be possible but not nice IMO. But from what I've seen there is also the `AliasFilingSubscriber`, which seems to do what we want.

Actually what we need is a combination of the `StructureTypeFilingBehavior` and the `AliasFilingBehavior`, but I guess they are currently not working together. Probably we should make that possible instead.

What do you think?

dantleech:
Well, there is already the `StructureTypeFilingSubscriber` in `sulu/sulu` - can you not just add the extra (alias) logic there? Then you _could_ make the `StructureTypeFilingSubscriber` extend the `AliasFilingSubscriber`, but it's such a simple change I wouldn't bother.

danrot:
Well, that would work, but using inheritance is not the cleanest solution IMO. Would be better if both Behaviors are implemented, so that both of them could be applied independent of each other.

And I think it should be possible quite easily, all we have to do is to register the subscribers with the correct priorities. The first one sets the base path, then there is another one adding the alias, and another one adding the structure type.

So the latter subscribers would look something like:

``` php
$path = $event->getParentNode()->getPath();
$subPath = some_magic();

$parentNode = $this->nodeManager->createPath($path . '/' . $subpath);

$event->setParentNode($parentNode);
```

Or does that have serious performance impacts?

dantleech:
I guess that would work, so the "Document Manager" subscriber would go
first, creating the document alias node if required, and the "Sulu"
subscriber would come next, adding the "Structure Type" node and setting
that as the parent. Right?

I think thats clean, I don't think the performance peanalty would be
signigicant, and it leaves the existing code untouched. Although I would
personally lean towards adding all the logic in the
StructureFilingSubscriber, as it's not that much code duplication and
means one less subscriber in the chain and that's always good. But
without looking closer at the problem I won't say either way.

So either of those solutions sound fine to me.

The way with the least code would be to extend the
AliasFilingSubscriber, but that's probably the worst solution in terms
of coupling.

danrot:
Yeah, you understood my suggestion correctly.

@stemod I would say we go with the approach I have suggested last, so that we have multiple subscribers extending the path step by step.

lib/Behavior/Path/ResetFilingPathBehavior.php
```diff
@@ -0,0 +1,27 @@
...
+/**
+ * Automatically positions the document at a configured location as a child of a node
+ * named after the documents alias.
+ *
+ * For example, if you specify the base path to be "/cmf/example" and the document has
+ * the alias "foobar" then the parent will be set to "/cmf/example/foobar".
+ *
+ * If the parent document does not exist, it will be created.
+ */
+interface ResetFilingPathBehavior extends ParentBehavior
```
danrot: I would say the interfaces `StructureTypeFilingSubscriber` and `AliasFilingSubscriber` should extend this interface, so that a document has not to implement this interface explicitly. Otherwise it would be possible to get the wrong behavior with the chaining of the path you had before.

lib/Behavior/Path/ResetFilingPathBehavior.php
```diff
@@ -0,0 +1,27 @@
...
+use Sulu\Component\DocumentManager\Behavior\Mapping\ParentBehavior;
+
+/**
+ * Automatically positions the document at a configured location as a child of a node
+ * named after the documents alias.
+ *
+ * For example, if you specify the base path to be "/cmf/example" and the document has
+ * the alias "foobar" then the parent will be set to "/cmf/example/foobar".
+ *
+ * If the parent document does not exist, it will be created.
```
danrot: This comment does not describe what this behavior is doing.

danrot:
@stemod You should also always have a look at the StyleCI check. It looks like they have changed some configuration, try to delete the `psr0` line in `.styleci.yml` and see what happens (if it rewrites some namespaces it is wrong)

lib/Behavior/Path/ResetFilingPathBehavior.php
```diff
@@ -0,0 +1,23 @@
...
+ */
+
+namespace Sulu\Component\DocumentManager\Behavior\Path;
+
+use Sulu\Component\DocumentManager\Behavior\Mapping\ParentBehavior;
+
+/**
+ * Resets the path to base path.
+ *
+ * this is used for example with the AliasFilingBehavior
```
danrot: I would omit this comment.

lib/Subscriber/Behavior/Path/ResetFilingPathSubscriber.php
```diff
@@ -0,0 +1,47 @@
+<?php
+
+namespace Sulu\Component\DocumentManager\Subscriber\Behavior\Path;
+
+use Sulu\Component\DocumentManager\Behavior\Path\ResetFilingPathBehavior;
+use Sulu\Component\DocumentManager\Event\PersistEvent;
+use Sulu\Component\DocumentManager\Events;
+
+/**
+ * Class ResetFilingPathSubscriber.
```
danrot: Please write a proper PHPDoc instead, with a bit of explanation.

danrot:
I am missing a test for the `ResetFilingSubscriber`. And I know it isn't your fault, but could you declare all the fields in the `AliasFilingSubscriberTest`, when you already touch it?

dantleech:
What does the `ResetFilingSubscriber` do? Why do we need it?

danrot:
@dantleech Since we chain the `AliasFilingSubscriber` and the `StructureFilingSubscriber` now, we need another Subscriber setting the base path. Instead of appending something to the path, as the other subscribers do, the `ResetFIlingSubscriber` sets the path absolutely. If we don't have this clear starting point the chaining of these subscribers is not possible.

dantleech:
This sounds wrong. Why does it need to be reset? Why is it not initialized as empty? Why can this not be part of an existing subscriber? Also this PR is based on the wrong branch I think.

danrot:
@dantleech Well, maybe `ResetFilingSubscriber` is a bad name... What we actually need is a subscriber which sets the absolute basepath (`/cmf` in the sulu case), but no other subscriber would fit for that. Since all the filing subscribers should be able to be chained in any way, you also can't put that logic to one of these.

Maybe it helps if we just rename it to `BasePathSubscriber`?

danrot:
And maybe it's not the best idea that this subscriber is inheriting from the `AbstractFilingSubscriber`, since it is not really doing filing, and the required methods are just implemented empty...

danrot:
Merged in 888c9d9

PR #43, opened 2015-10-02
Fixed undefined ghost title

wachterjohannes:
Often in sulu the title of ghost pages is undefined. The proxy-factory uses different options (`[]`) to load childs then parent.

**tasks:**
- [x] test coverage

**informations:**

| q | a |
| --- | --- |
| Fixed tickets | none |
| BC breaks | none |
| Documentation PR | none |

### DIFF
--- a/lib/Subscriber/Core/RegistratorSubscriber.php
+++ b/lib/Subscriber/Core/RegistratorSubscriber.php
@@ -211,8 +211,8 @@ class RegistratorSubscriber implements EventSubscriberInterface
         $this->documentRegistry->clear();
     }
 
-    /*
-     * Register the document and apparently update the locale --
+    /**
+     * Register the document and apparently update the locale --.
      *
      * TODO: Is locale handling already done above??
      *
--- a/tests/Unit/ProxyFactoryTest.php
+++ b/tests/Unit/ProxyFactoryTest.php
@@ -12,19 +12,24 @@
 namespace Sulu\Component\DocumentManager\Tests\Unit;
 
 use PHPCR\NodeInterface;
+use Prophecy\Argument;
 use ProxyManager\Factory\LazyLoadingGhostFactory;
 use ProxyManager\Proxy\LazyLoadingInterface;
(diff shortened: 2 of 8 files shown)


wachterjohannes:
@dantleech this PR should be good if the tests will be green (:

lib/Subscriber/Core/RegistratorSubscriber.php
```diff
@@ -218,6 +218,7 @@ public function handleClear(ClearEvent $event)
      *
      * @param AbstractMappingEvent $event
      */
+
```
dantleech: Something is wrong with StyleCI.

wachterjohannes: no the phpdoc is wrong (:

dantleech:
This seems to make sense, just to be clear:
- Previously the parent was created as a proxy but without options.
- This meant that if "load_ghost_content" was `true` when fetching a document, it's parent would have this value set to `false.`
- With this change the parent inherits the original documents options, so if a document is loaded with `load_ghost_content` as `true`, the parent will also have this option.

I wonder what would happen if you called `find` on the same object with a different set of options?

``` php
$dm->find('/foobar', ["bar" => "foo"]); 
$dm->find('/foobar', '["bar" => "bar"]);`
```

Would the parent have the `bar` value or the `foo` value?

wachterjohannes:
hmm i cannt say but i think he has the second one because it will be rehydrated?

dantleech:
In theory it should be.. maybe just worth remembering in case it's an issue.

tests/Unit/ProxyFactoryTest.php
```diff
@@ -61,6 +66,36 @@ public function testCreateProxy()
     }
 
     /**
+     * It should populate the documents parent property (with custom options) with a proxy.
```
dantleech, line 69: `document's` (with an apostrophe) because its about possesion - an `s` without an apostrophe just means plurality.

dantleech:
Merged, we can fix the apostrophe later :)

Issue #41, opened 2015-10-01
Wrong document for locale


wachterjohannes:
if you load a document with any locale this document will be stored in the `DocumentRegistry` with this locale.
If you load in the same request the same uuid as document with a different locale you get a document with the first locale. this seems to be a bug in the `DocumentRegistry`.

Because the `RegistratorSubscriber` ignores the given locale and loads any document out of the registry.

My Suggestion would be to change this behavior to this: https://github.com/sulu-io/sulu-document-manager/pull/42

wachterjohannes:
@dantleech what do you think? /cc @danrot

dantleech:
I don't think this is a bug with the DocumentRegistry.

The Document is the object. The essential role of the document registry is to return the object. If an object is loaded in a different locale, it is stil the same object. The DocumentRegistry doesn't actually care what is _in_ the object.

The `DocumentManager` does, perhaps confusingly, also track the current locale of the object. But this is a separate responsibility.

What you describe should not happen because the locale should be updated (and the document _rehydrated_) if the locale is [different from the currently registered locale for the document](https://github.com/sulu-io/sulu-document-manager/blob/query_builder_final/lib/Subscriber/Core/RegistratorSubscriber.php#L109).

If you could add a failing test I could perhaps look at this if you cannot find it.

wachterjohannes:
i will take a look today ... but for me its not the same document because the locale is a big part of it and should also be tracked in the registry! but i will take a look at a testcase today!

wachterjohannes:
hmm it is not a simple `DocumentManager` problem. because it seams to be a problem in combination with sulu´s `StructureSubscriber`.

But for me it should be a different Object for "en" or "de". Because if a bigger routine get a document for "en" and a sub-routing uses the same uuid to get the document in "de" the bigger routine did not recognize that his document has changed and the locale of it is different. so i think the locale should also be used to determine the document in the Registry. what do you think @dantleech @danrot ? i have created a failing testcase in PR #44

dantleech:
The document and the object are the same thing. You can look at the object in different ways (e.g. with differnet locales) but essentially it is the same object representing the same subject. The object ID maps directly to the node UUID. If you had lots of different representations of the document floating about it would become difficult to manage.

One thing that springs to mind is that if you have 10 objects representing one document (i.e. 10 locales), each one had 3 translated fields and 3 non-translated fields, how should conflicts be resolved if one of them changes a non-translated field?

[...]

PR #44, opened 2015-10-02
Fixed document different locale

wachterjohannes:
For different locales the document manager did not return different objects. so you can overwrite properties of a document which could be used somewhere else without knowing ...

The easiest way would be to handle the locale in the `DocumentRegistry`. there is one document for a node in english and one in german for example. 

**tasks:**
- [ ] test coverage
- [ ] submit changes to the documentation

**informations:**

| q | a |
| --- | --- |
| Fixed tickets | #41 |
| BC breaks | none |
| Documentation PR | none |

### DIFF
--- a/tests/Functional/DocumentManager/FindTest.php
+++ b/tests/Functional/DocumentManager/FindTest.php
@@ -74,4 +74,26 @@ class FindTest extends BaseTestCase
         $document = $persistedDocument;
         $this->assertEquals('en', $document->getLocale());
     }
+
+    /**
+     * It can load different locales from the same document in the same request.
+     */
+    public function testFindMultipleDifferentLocales()
+    {
+        $this->generateDataSet([
+            'locales' => ['en', 'de'],
+        ]);
+
+        $manager = $this->getDocumentManager();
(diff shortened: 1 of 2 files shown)


tests/Functional/DocumentManager/FindTest.php
```diff
@@ -74,4 +74,26 @@ public function testPersistFindNoLocales()
...
+        $manager = $this->getDocumentManager();
+        $manager->flush();
+
+        $result1 = $manager->find(self::BASE_PATH, 'en');
+        $this->assertEquals('en', $result1->getLocale());
+
+        $result2 = $manager->find(self::BASE_PATH, 'de');
+        $this->assertEquals('de', $result2->getLocale());
+
+        $this->assertNotEquals(false, $result1 === $result2);
```
dantleech, line 96: You could better use:

``` php
$this->assertSame($result1, $result2);
```

dantleech:
Closing this, as the current behavior is correct, calling `find()` will change the instances locale. Feel free to reopen if you think this tests adds value anyway.

Issue #45, opened 2015-10-03
Options and global state


dantleech:
Options should be treated as locales .. if a document is requested with different options, then it should be rehydrated. It's relationship with other nodes should be reset (so that they will be hydrated with the same options upon request).

If `find` is called without options (i.e. `$options = null`) then the the loaded, hydrated, documetn should be returneded without an additional hydration request.

This feature should be done at the same time as making `$locale` an option, which would simplify things - locale is effectively another option, albeit one which is currently built into the fundamentals of the system.

The problem with this approach is that requesting a document with a different set of options changes the state of not just the requested document, but all documents associated with it (lazily or not).

Ideas:
- Add the possiblity to retrieve immutable clones of a document instead of the "real" document (e.g. an optoin `immutable` => true`).
- Always return new objects and resolve conflicts that arrise by applying changes from multiple versions in the order in which they were requested. (conflicts would arrise if non-translated fields, for example, where modified on multiple versions of the same document).
- We simply say that users of the document manager are responsibule for ensuring that their documents are in the desired state.

Other ideas welcome.. 

/cc @wachterjohannes

danrot:
I'd prefer the second option, because it is the most comfortable one for the user of this library.

A problem I can see is that there are also options passed, which would not necessarily cause a rehydration, so it would have an unnecessary performance impact.

wachterjohannes:
hmm at the moment for me the option 3 would be perfect ... but later we should look how to handle conflicts and solve them! we get already problems with multiple localed documents ... which are very weird (: we should be able to hydrate the same node with different locales in different nodes!

danrot:
Letting the user handle this on his own doesn't sound like the perfect solution for me... That's just error-prone...

PR #46, opened 2015-10-03
Rehydrate option

dantleech:
Added an option to disable "rehydration" of a document which is already in the document registry.

The behavior is as follows:
- If `rehydrate` is false, then `RegistationSubscriber::handleDocumentFromRegistry` method will stop further event propagation if the document is already in the registry.

This allows the document to be retrieved from the document manager without changing it's global state.

### DIFF
--- a/lib/Subscriber/Core/RegistratorSubscriber.php
+++ b/lib/Subscriber/Core/RegistratorSubscriber.php
@@ -14,13 +14,13 @@ namespace Sulu\Component\DocumentManager\Subscriber\Core;
 use Sulu\Component\DocumentManager\DocumentRegistry;
 use Sulu\Component\DocumentManager\Event\AbstractMappingEvent;
 use Sulu\Component\DocumentManager\Event\ClearEvent;
+use Sulu\Component\DocumentManager\Event\ConfigureOptionsEvent;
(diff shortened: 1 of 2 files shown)


lib/Subscriber/Core/RegistratorSubscriber.php
```diff
@@ -98,6 +108,15 @@ public function handleDocumentFromRegistry(HydrateEvent $event)
         $document = $this->documentRegistry->getDocumentForNode($node);
 
         $event->setDocument($document);
+
+        $options = $event->getOptions();
+
+        // if reydration is not required (f.e. we just want to retrieve the
+        // current state of the document, no matter it's current state) stop
+        // further event propagation - we have the document now.
+        if (isset($options['rehydrate']) && false === $options['rehydrate']) {
```
dantleech, line 120: Note that the forced use of `isset` is a bug. Caused by the fact that options are not inherited by proxies, making it possible for an empty array to be passed as `$options`. This is partially addressed by @wachterjohannes 's fix, but not completely apparently. I think there is a better way, see #45

wachterjohannes:
@dantleech i will try this on monday!

wachterjohannes:
@dantleech this seems to help (: thanks for your help

PR #49, opened 2015-10-14
added options to RemoveEvent

danrot:
Required for the dereference option described in https://github.com/sulu-io/sulu/issues/877

**tasks:**
- [ ] test coverage
- [ ] submit changes to the documentation

**informations:**

| q | a |
| --- | --- |
| Fixed tickets | none |
| BC breaks | none |
| Documentation PR | TODO |

### DIFF
--- a/lib/Exception/DocumentReferencedException.php
+++ b/lib/Exception/DocumentReferencedException.php
@@ -1,0 +1,69 @@
+<?php
+/*
+ * This file is part of Sulu.
+ *
+ * (c) MASSIVE ART WebServices GmbH
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace Sulu\Component\DocumentManager\Exception;
+
+use PHPCR\PropertyInterface;
+
+/**
+ * This is exception is thrown if a document is still referenced, and the executed action does not work because of that.
+ */
+class DocumentReferencedException extends DocumentManagerException
+{
+    /**
+     * @var object
+     */
+    private $document;
+
+    /**
+     * @var PropertyInterface[]
+     */
+    private $references;
+
+    /**
+     * @param object $document
+     * @param PropertyInterface[] $references
+     * @param \Exception $previous
+     */
+    public function __construct($document, array $references = [], \Exception $previous = null)
+    {
+        $this->document = $document;
+        $this->references = $references;
+
+        parent::__construct(
+            sprintf(
+                'The document with the object hash "%s" cannot be deleted because it is still referenced',
+                spl_object_hash($this->document)
+            ),
+            0,
+            $previous
+        );
+    }
+
+    /**
+     * Returns the document which cannot be deleted because it is referenced.
+     *
+     * @return object
+     */
+    public function getDocument()
+    {
+        return $this->document;
+    }
+
+    /**
+     * Returns all the references causing the page not to be deletable.
+     *
+     * @return PropertyInterface[]
+     */
+    public function getReferences()
+    {
+        return $this->references;
+    }
+}
--- a/lib/Subscriber/Phpcr/RemoveSubscriber.php
+++ b/lib/Subscriber/Phpcr/RemoveSubscriber.php
@@ -11,10 +11,11 @@
 
 namespace Sulu\Component\DocumentManager\Subscriber\Phpcr;
 
+use PHPCR\PropertyInterface;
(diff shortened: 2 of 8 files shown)


lib/Exception/DocumentReferencedException.php
```diff
@@ -0,0 +1,69 @@
...
+     * @param \Exception $previous
+     */
+    public function __construct($document, array $references = [], \Exception $previous = null)
+    {
+        $this->document = $document;
+        $this->references = $references;
+
+        parent::__construct(
+            sprintf(
+                'The document with the object hash "%s" cannot be deleted because it is still referenced',
```
dantleech, line 42: Would prefer this to be an inline message, as with all the other exception messages in this libary:

``` php
throw new BlahException(sprintf('This is the message'));
```

As we have discussed before. It is clearer to read in the source code, and usually each instance that an exception is thrown has a slightly different context - usually information can be given to help the developer solve their problems.

If the exception is only used in one place, then putting the message in the Exception class just means that information (i.e. explanation for why exception would be thrown) that would be available when reading the source has to be looked up in a new class.

I would prefer, regardless of other policies, that we maintain the current standard in this library.

lib/Subscriber/Phpcr/RemoveSubscriber.php
```diff
@@ -55,12 +56,20 @@ public static function getSubscribedEvents()
...
      * @param RemoveEvent $event
+     *
+     * @throws DocumentReferencedException
      */
     public function handleRemove(RemoveEvent $event)
     {
         $document = $event->getDocument();
         $node = $this->documentRegistry->getNodeForDocument($document);
 
+        $references = iterator_to_array($node->getReferences());
```
dantleech, line 70: This should not be done here. It adds a potentially very significant amount of overhead. The implementation, i.e. Jackalope, should handle this, not us.

This makes doing things in batches potentially slow. This should be handled in Sulu AFTER the PHPCR implementation has thrown a Referential integrity exception. Failing that, it should be done in the consuming code (Sulu).

danrot, line 70: I tried to do that, but it was not possible due to a few issues in jackalope, and some architectural decisions you made in the document manager (e.g. the abstinence of a unit of work). The exception is not thrown before the `flush` event, and I found no way to get the affected nodes at this spot.

danrot, line 70: Just did a quick search, the only thing I could do is using `iterator_count` here, don't know how much of a difference this is in terms of performance.

lib/Exception/DocumentReferencedException.php
danrot, line 42: I have never agreed to that :wink: Apart from that, I I really need to pass the document and references here. In addition I ran into issues, in exceptions following your pattern, when I actually wanted to react on them accordingly in a catch block.

danrot:
Related issue in jackalope: https://github.com/jackalope/jackalope/issues/298

lib/Exception/DocumentReferencedException.php
dantleech, line 42: I know, but that is what is done in this library. You can still throw a typed exception, but I just prefer that the message is inline. I have seen instances where people just throw an existing exception that "looks like the right one" but gives a misleading error message. The message is the single most important thing of an Exception most of the time, it explains what has gone wrong and good exceptions also tell the developer how to fix the issue.

lib/Subscriber/Phpcr/RemoveSubscriber.php
dantleech, line 70: My point is that this should be done in Sulu. When somebody issues a "remove" command to the document manager it should do the minmum required to fulfil that request. Anything else is just a obligatory overhead on all requests, ignoring if they need this functionality or not.

I would rather this was done in the `sulu/sulu` library than here, otherwise we could add an option to explicitly command the DM to do this, but I would like that to be discussed more. i think it is important to keep the DM as minimal as possible.

danrot, line 70: Well, I guess we could move that check to sulu/sulu, but IMO that's an issue also the document manager should be caring about.

However, because of some bugs in jackalope I can't continue to fix this the proper way, so this PR will stay open. Have fixed it in a less cleaner way now, and created an issue for later processing: https://github.com/sulu-io/sulu/issues/1710

PR #52, opened 2015-11-04
Fixed empty locale when generating property names

### LINKED ISSUE #1704: Properties without locale in database
I have a lot of pages with properties (in phpcr) which have a name like this: `i18n:-excerpt-title` i think it only happens with extensions but i am not quite sure.

wachterjohannes:
**tasks:**
- [x] test coverage

**informations:**

| q | a |
| --- | --- |
| Fixed tickets | fixes https://github.com/sulu-io/sulu/issues/1704 |
| BC breaks | none |
| Documentation PR | none |

### DIFF
--- a/lib/Subscriber/Core/MappingSubscriber.php
+++ b/lib/Subscriber/Core/MappingSubscriber.php
@@ -102,8 +102,13 @@ class MappingSubscriber implements EventSubscriberInterface
      * @param mixed $locale
      * @param mixed $fieldMapping
      */
-    private function persistReference(NodeInterface $node, DocumentAccessor $accessor, $fieldName, $locale, $fieldMapping)
-    {
+    private function persistReference(
+        NodeInterface $node,
+        DocumentAccessor $accessor,
+        $fieldName,
+        $locale,
+        $fieldMapping
+    ) {
         $referenceDocument = $accessor->get($fieldName);
 
         if (!$referenceDocument) {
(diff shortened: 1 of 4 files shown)


lib/Subscriber/Core/MappingSubscriber.php
```diff
@@ -111,14 +116,21 @@ private function persistReference(NodeInterface $node, DocumentAccessor $accesso
...
 
         $referenceNode = $this->documentRegistry->getNodeForDocument($referenceDocument);
-        $phpcrName = $this->encoder->encode($fieldMapping['encoding'], $fieldMapping['property'], $locale);
+        try {
+            $phpcrName = $this->encoder->encode($fieldMapping['encoding'], $fieldMapping['property'], $locale);
+        } catch (\InvalidArgumentException $ex) {
+            // arguments unvalid no valid propertyname could be generated (e.g. no locale given for localized encoding)
+            return;
+        }
         $node->setProperty($phpcrName, $referenceNode);
```
danrot: Why don't you just move that also into the try block, and therefore remove the `return` statement in the catch block? I think we said we either return early or at the very end, that's somewhere in between.

lib/Subscriber/Core/MappingSubscriber.php
```diff
@@ -131,9 +143,19 @@ private function persistReference(NodeInterface $node, DocumentAccessor $accesso
...
+        DocumentAccessor $accessor,
+        $fieldName,
+        $locale,
+        array $fieldMapping
+    ) {
+        try {
+            $phpcrName = $this->encoder->encode($fieldMapping['encoding'], $fieldMapping['property'], $locale);
+        } catch (\InvalidArgumentException $ex) {
+            // arguments unvalid no valid propertyname could be generated (e.g. no locale given for localized encoding)
+            return;
```
danrot, line 161: Same as above

lib/Subscriber/Core/MappingSubscriber.php
```diff
@@ -184,10 +206,22 @@ public function handleHydrate(AbstractMappingEvent $event)
...
```
danrot, line 233: Same as above

lib/Subscriber/Core/MappingSubscriber.php
```diff
@@ -208,9 +242,19 @@ private function hydrateReferenceField(NodeInterface $node, $document, DocumentA
...
```
danrot, line 262: Same as above

danrot:
Do we need the catch blocks at all? Shouldn't these exceptions never be thrown, if all the subscribers are working correctly?

wachterjohannes:
this is an own subscriber which can work without locale (unlocalized properties). so it has to be checked here.

wachterjohannes:
@danrot fixed

wachterjohannes:
@danrot fixed

danrot:
Shouldn't that PR be targeted at develop?

wachterjohannes:
yes i will change this

PR #58, opened 2016-02-01
Added missing features for custom-url feature

wachterjohannes:
Mein PR https://github.com/sulu-io/sulu/pull/1927

### DIFF
--- a/lib/PathSegmentRegistry.php
+++ b/lib/PathSegmentRegistry.php
@@ -55,8 +55,6 @@ class PathSegmentRegistry
             );
         }
 
-        $name = $this->pathSegments[$name];
-
-        return $name;
+        return $this->pathSegments[$name];
     }
 }
--- a/lib/Subscriber/Behavior/Audit/TimestampSubscriber.php
+++ b/lib/Subscriber/Behavior/Audit/TimestampSubscriber.php
@@ -11,14 +11,15 @@
 
 namespace Sulu\Component\DocumentManager\Subscriber\Behavior\Audit;
 
+use Sulu\Component\DocumentManager\Behavior\Audit\LocalizedTimestampBehavior;
 use Sulu\Component\DocumentManager\Behavior\Audit\TimestampBehavior;
+use Sulu\Component\DocumentManager\Event\MetadataLoadEvent;
 use Sulu\Component\DocumentManager\Event\PersistEvent;
 use Sulu\Component\DocumentManager\Events;
 use Symfony\Component\EventDispatcher\EventSubscriberInterface;
 
 /**
- * Manage the timestamp (created, changed) fields on
- * documents before they are persisted.
+ * Manage the timestamp (created, changed) fields on documents before they are persisted.
  */
 class TimestampSubscriber implements EventSubscriberInterface
 {
@@ -37,26 +38,31 @@ class TimestampSubscriber implements EventSubscriberInterface
     }
 
     /**
-     * @param MetadataLoadEvent
+     * @param MetadataLoadEvent $event
      */
-    public function handleMetadataLoad($event)
+    public function handleMetadataLoad(MetadataLoadEvent $event)
     {
-        if (!$event->getMetadata()->getReflectionClass()->isSubclassOf(TimestampBehavior::class)) {
+        if (!$event->getMetadata()->getReflectionClass()->isSubclassOf(LocalizedTimestampBehavior::class)) {
             return;
+        }
+
+        $encoding = 'system_localized';
+        if ($event->getMetadata()->getReflectionClass()->isSubclassOf(TimestampBehavior::class)) {
+            $encoding = 'system';
         }
 
         $metadata = $event->getMetadata();
         $metadata->addFieldMapping(
             self::CREATED,
             [
-                'encoding' => 'system_localized',
+                'encoding' => $encoding,
                 'property' => self::CREATED,
             ]
         );
         $metadata->addFieldMapping(
             self::CHANGED,
             [
-                'encoding' => 'system_localized',
+                'encoding' => $encoding,
                 'property' => self::CHANGED,
             ]
         );
@@ -69,7 +75,7 @@ class TimestampSubscriber implements EventSubscriberInterface
     {
         $document = $event->getDocument();
 
-        if (!$document instanceof TimestampBehavior) {
+        if (!$document instanceof LocalizedTimestampBehavior) {
             return;
         }
 
--- a/lib/Subscriber/Core/MappingSubscriber.php
+++ b/lib/Subscriber/Core/MappingSubscriber.php
@@ -142,7 +142,7 @@ class MappingSubscriber implements EventSubscriberInterface
             $phpcrName = $this->encoder->encode($fieldMapping['encoding'], $fieldMapping['property'], $locale);
             $node->setProperty($phpcrName, $referenceNode);
         } catch (InvalidLocaleException $ex) {
-            // arguments unvalid no valid propertyname could be generated (e.g. no locale given for localized encoding)
+            // arguments invalid, no valid propertyname could be generated (e.g. no locale given for localized encoding)
             return;
         }
     }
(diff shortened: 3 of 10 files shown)


lib/Behavior/Audit/NonLocalizedTimestampBehavior.php
```diff
@@ -0,0 +1,18 @@
...
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace Sulu\Component\DocumentManager\Behavior\Audit;
+
+/**
+ * Adds the date when created and lastly changed the document (without locale).
+ */
+interface NonLocalizedTimestampBehavior extends TimestampBehavior
```
danrot: Would prefer this to be called `TimestampBehavior` and the other interface should be the `LocalizedTimestampBehavior`. Negations in class names make things more complicated to understand... I think we don't have a BC promise here, do we? Then I would definitely rename this, otherwise we'll have this debt for a long time...

lib/PathSegmentRegistry.php
```diff
@@ -35,6 +35,17 @@ public function __construct(array $pathSegments = [])
     }
 
     /**
+     * Add a pathsegment with given name.
+     *
+     * @param string $name
+     * @param string $pathSegment
+     */
+    public function add($name, $pathSegment)
```
danrot: If the suggestion in https://github.com/sulu-io/sulu-document-manager/pull/58/files#diff-aaf1b8253014c99a5f92c8b8de4d9515R43 would be implemented, we don't even need this method.

lib/Subscriber/Behavior/Audit/TimestampSubscriber.php
```diff
@@ -45,18 +46,23 @@ public function handleMetadataLoad($event)
             return;
         }
 
+        $encoding = 'system_localized';
+        if ($event->getMetadata()->getReflectionClass()->isSubclassOf(NonLocalizedTimestampBehavior::class)) {
+            $encoding = 'system';
+        }
```
danrot, line 52: This would be the first subscriber not being directly related to a single Behavior (this was quite nice until now IMO, because it was always clear where to look for the operations executed because of these behaviors).

@dantleech Do you have any better ideas? For me it would be even better to use inheritance here (an `AbstractTimestampBehavior` which is the base class for a `TimestampBehavior` and a `LocalizedTimestampBehavior`, whereby the derived classes only differ in their usage of the encoding property).

lib/Subscriber/Core/MappingSubscriber.php
```diff
@@ -172,6 +175,33 @@ private function persistGeneric(
...
+        $locale,
+        array $fieldMapping
+    ) {
+        try {
+            $phpcrName = $this->encoder->encode($fieldMapping['encoding'], $fieldMapping['property'], $locale);
+            $value = $accessor->get($fieldName);
+            $this->validateFieldValue($value, $fieldName, $fieldMapping);
+            $node->setProperty($phpcrName, json_encode($value));
+        } catch (InvalidLocaleException $ex) {
+            // arguments unvalid no valid propertyname could be generated (e.g. no locale given for localized encoding)
```
danrot: `arguments invalid, no valid ...`

lib/Subscriber/Core/MappingSubscriber.php
```diff
@@ -222,7 +264,8 @@ private function hydrateReferenceField(
         DocumentAccessor $accessor,
         $fieldName,
         $locale,
-        array $fieldMapping
+        array $fieldMapping,
+        array $options
```
danrot, line 268: For which options was this required?

danrot:
CHANGELOG missing.

lib/Behavior/Audit/NonLocalizedTimestampBehavior.php
webmozart: :+1:

lib/Subscriber/Behavior/Audit/TimestampSubscriber.php
dantleech, line 52: You could implement the logic in the CustomUrlSubscriber (assuming it exists).

I am currently doing something similar as this, but checking for the `LocaleBehavior`, which doesn't make much more sense, but would at least avoid another interface.

I think the best solution would be to make the "localized" status a property of the Metadata, so, in a subscriber for the `CustomUrlDocument`:

```
$metadata = $event->getMetadata();
$metadata->setLocalized(true);
```

or in TAML

```
Some/Bundle/Document/CustomUrlDocument
    # ...
    localized: true
```

Then you would be able to access this property from the timestampable behavior and in other places.

I can make it so.

tests/Unit/PathSegmentRegistryTest.php
```diff
@@ -24,12 +29,22 @@ public function setUp()
     }
 
     /**
+     * It should add and retrieve a path segment.
+     */
+    public function testAdd()
+    {
+        $this->pathRegistry->add('test', 'test-segment');
+        $segment = $this->pathRegistry->getPathSegment('test');
+        self::assertEquals('test-segment', $segment);
```
dantleech: $this->assertEquals ..

tests/Unit/PathSegmentRegistryTest.php
```diff
@@ -24,12 +29,22 @@ public function setUp()
...
+    }
+
+    /**
      * It should retrieve a path segment.
      */
     public function testGetPathSegment()
     {
         $segment = $this->pathRegistry->getPathSegment('base');
-        $this->assertEquals('cmf', $segment);
+        self::assertEquals('cmf', $segment);
```
dantleech: $this

lib/Subscriber/Core/MappingSubscriber.php
```diff
@@ -97,6 +97,9 @@ public function handlePersist(PersistEvent $event)
                 case 'reference':
                     $this->persistReference($node, $accessor, $fieldName, $locale, $fieldMapping);
                     break;
+                case 'json_array':
+                    $this->persistJsonArray($node, $accessor, $fieldName, $locale, $fieldMapping);
+                    break;
```
dantleech, line 102: We need to add validation for types, created: https://github.com/sulu-io/sulu-document-manager/issues/68

lib/Subscriber/Core/MappingSubscriber.php
dantleech: Catching this exception and silently failing will cause unexpected behavior.

lib/Subscriber/Core/MappingSubscriber.php
```diff
@@ -272,6 +315,35 @@ private function hydrateGenericField(
...
+    ) {
+        try {
+            $phpcrName = $this->encoder->encode($fieldMapping['encoding'], $fieldMapping['property'], $locale);
+            $value = $node->getPropertyValueWithDefault(
+                $phpcrName,
+                $this->getDefaultValue($fieldMapping)
+            );
+            $accessor->set($fieldName, json_decode($value, true));
+        } catch (InvalidLocaleException $ex) {
+            // arguments unvalid no valid propertyname could be generated (e.g. no locale given for localized encoding)
```
dantleech: If no locale was given, then that we should probably be using the default locale.. would prefer if we removed these Catches.

lib/Subscriber/Core/MappingSubscriber.php
dantleech, line 268: I think we should always propagate the options to the proxy. This was likely an existing bug /cc @wachterjohannes

lib/Subscriber/Behavior/Audit/TimestampSubscriber.php
wachterjohannes, line 52: fixed with one subscriber

lib/Subscriber/Core/MappingSubscriber.php
wachterjohannes: but this is done for all "types" in this class maybe add an issue? typo is fixed

lib/Subscriber/Core/MappingSubscriber.php
wachterjohannes, line 268: @dantleech is right this was missing

lib/Subscriber/Core/MappingSubscriber.php
wachterjohannes: like i said this is "copied" from the other types we should add an issue and fix that in another PR

wachterjohannes:
@danrot should i introduce a UPGRADE here?

dantleech:
Created https://github.com/sulu-io/sulu-document-manager/issues/70.

PR #60, opened 2016-02-12
Query Builder

dantleech:
Introduces a QueryBuilder based on PHPCR-ODM

### DIFF
--- a/lib/DocumentStrategyInterface.php
+++ b/lib/DocumentStrategyInterface.php
@@ -12,9 +12,11 @@
 namespace Sulu\Component\DocumentManager;
 
 use PHPCR\NodeInterface;
+use PHPCR\Query\QOM\QueryObjectModelFactoryInterface;
 
 /**
- * Document strategies determine how documents are managed.
+ * Document strategies determine how documents are managed, for example
+ * if the document class should be determined by a mixin, or by the primary node type.
  */
 interface DocumentStrategyInterface
 {
@@ -48,4 +50,48 @@ interface DocumentStrategyInterface
      * @return Metadata|null
      */
     public function resolveMetadataForNode(NodeInterface $node);
+
+    /**
+     * Return the primary node type from the given document class FQN.
+     *
+     * For example:
+     *
+     * ```
+     * return 'nt:unstructured';
+     * ```
+     *
+     * @param string $classFqn
+     *
+     * @return string
+     */
+    public function getPrimaryNodeType($classFqn);
+
+    /**
+     * Create a source constraint for a source document. That is return the
+     * constraint that should be used to return only documents of the given
+     * class FQN.
+     *
+     * For example:
+     *
+     * ```
+     * return $qomf->comparison(
+     *     $qomf->propertyValue(
+     *         $sourceNode->getAlias(),
+     *         'jcr:mixinTypes'
+     *     ),
+     *     QOMConstants::JCR_OPERATOR_EQUAL_TO,
+     *     $qomf->literal('foo')
+     * );
+     * ```
+     *
+     * Can return NULL if no constraints should be added. This may be required
+     * if the primary type already represents the document class.
+     *
+     * @param QueryObjectModelFactoryInterface $qomf
+     * @param string $sourceAlias
+     * @param string $classFqn
+     *
+     * @return \PHPCR\Query\QOM\ConstraintInterface
+     */
+    public function createSourceConstraint(QueryObjectModelFactoryInterface $qomf, $sourceAlias, $classFqn);
 }
--- a/lib/Metadata.php
+++ b/lib/Metadata.php
@@ -36,17 +36,17 @@ class Metadata
     /**
      * @var array
      */
-    private $fieldMappings;
+    private $fieldMappings = [];
 
     /**
      * Add a field mapping for field with given name, for example:.
      *
-     * ````
+     * ```
      * $metadata->addFieldMapping(array(
      *     'encoding' => 'content',
      *     'property' => 'phpcr_property_name',
      * ));
-     * ````
+     * ```
      *
      * @param string $name Name of field/property in the mapped class.
      * @param array $mapping {
@@ -82,6 +82,23 @@ class Metadata
     }
 
     /**
+     * Return the mapping for the named field.
+     *
+     * @return array
+     */
+    public function getFieldMapping($name)
+    {
+        if (!isset($this->fieldMappings[$name])) {
+            throw new \InvalidArgumentException(sprintf(
+                'Field mapping "%s" not known in metadata for document type "%s". Known mappings: "%s"',
+                $name, $this->alias, implode('", "', array_keys($this->fieldMappings))
+            ));
+        }
+
+        return $this->fieldMappings[$name];
+    }
+
+    /**
      * @return string
      */
     public function getClass()
@@ -99,7 +116,7 @@ class Metadata
     }
 
     /**
-     * @return ReflectionClass
+     * @return \ReflectionClass
      */
     public function getReflectionClass()
     {
--- a/lib/Query/QueryBuilderConverter.php
+++ b/lib/Query/QueryBuilderConverter.php
@@ -1,0 +1,256 @@
+<?php
+
+/*
+ * This file is part of Sulu.
+ *
+ * (c) MASSIVE ART WebServices GmbH
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace Sulu\Component\DocumentManager\Query;
+
+use Doctrine\ODM\PHPCR\Query\Builder\AbstractNode as QBConstants;
+use Doctrine\ODM\PHPCR\Query\Builder\ConverterPhpcr;
+use Doctrine\ODM\PHPCR\Query\Builder\OperandDynamicField;
+use Doctrine\ODM\PHPCR\Query\Builder\QueryBuilder;
+use Doctrine\ODM\PHPCR\Query\Builder\SourceDocument;
+use PHPCR\SessionInterface;
+use Sulu\Component\DocumentManager\DocumentStrategyInterface;
+use Sulu\Component\DocumentManager\Metadata;
+use Sulu\Component\DocumentManager\MetadataFactoryInterface;
+use Sulu\Component\DocumentManager\PropertyEncoder;
+use Symfony\Component\EventDispatcher\EventDispatcherInterface;
+
+/**
+ * Class which converts a Builder tree to a PHPCR Query.
+ */
+class QueryBuilderConverter extends ConverterPhpcr
+{
+    /**
+     * @var EventDispatcherInterface
+     */
+    private $eventDispatcher;
+
+    /**
+     * @var Metadata[]
+     */
+    protected $documentMetadata = [];
+
+    /**
+     * @param PropertyEncoder
+     */
+    protected $encoder;
+
+    /**
+     * @param DocumentStrategyInterface
+     */
+    private $strategy;
+
+    /**
+     * @param MetadataFactoryInterface
+     */
+    private $metadataFactory;
+
+    public function __construct(
+        SessionInterface $session,
+        EventDispatcherInterface $eventDispatcher,
+        MetadataFactoryInterface $metadataFactory,
+        PropertyEncoder $encoder,
+        DocumentStrategyInterface $strategy
+    ) {
+        $this->eventDispatcher = $eventDispatcher;
+        $this->qomf = $session->getWorkspace()->getQueryManager()->getQOMFactory();
+        $this->metadataFactory = $metadataFactory;
+        $this->encoder = $encoder;
+        $this->strategy = $strategy;
+    }
+
+    /**
+     * {@inheritdoc}
+     *
+     * @return Query
+     */
+    public function getQuery(QueryBuilder $builder)
+    {
+        $this->documentMetadata = [];
+        $this->sourceDocumentNodes = [];
+        $this->constraint = null;
+
+        $this->locale = $builder->getLocale();
+
+        if (!$this->locale) {
+            throw new \InvalidArgumentException(sprintf(
+                'No locale specified'
+            ));
+        }
+
+        $from = $builder->getChildrenOfType(QBConstants::NT_FROM);
+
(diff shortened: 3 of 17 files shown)


danrot:
Changelog is missing.

lib/Metadata.php
```diff
@@ -82,6 +82,25 @@ public function getFieldMappings()
     }
 
     /**
+     * Return the mapping for the named field.
+     *
+     * @see Metadata::addFieldMapping
```
danrot: param is missing here

danrot: And what is this `@see` about? You are referencing to the method itself here, aren't you?

lib/Query/QueryBuilderConverter.php
```diff
@@ -0,0 +1,256 @@
...
+
+        if (!$this->locale) {
+            throw new \InvalidArgumentException(sprintf(
+                'No locale specified'
+            ));
+        }
+
+        $from = $builder->getChildrenOfType(
+            QBConstants::NT_FROM
+        );
```
danrot: Wouldn't split the lines here, since it is not necessary.

lib/Query/QueryBuilderConverter.php
```diff
@@ -0,0 +1,256 @@
...
+
+        list($alias, $phpcrProperty) = $this->getPhpcrProperty(
+            $alias,
+            $field
+        );
+
+        $operand = $this->qomf->propertyValue(
+            $alias,
+            $phpcrProperty
+        );
```
danrot: Same as above, splitting lines here is probably not necessary.

lib/Query/QueryBuilderConverter.php
```diff
@@ -0,0 +1,256 @@
...
+     * field name and query alias.
+     *
+     * @param string $alias As specified in the query source.
+     * @param string $field Name of the document field
+     *
+     * @return array {
+     *
+     *     @var string Element is the real alias to use, second element is
+     *     @var string the property name
+     * }
```
danrot, line 218: Do these annotations really exist like this?

lib/Query/QueryBuilderConverter.php
```diff
@@ -0,0 +1,256 @@
...
+        $this->applySourceConstraints($builder);
+
+        $phpcrQuery = $this->qomf->createQuery(
+            $this->from,
+            $this->constraint,
+            $this->orderings,
+            $this->columns
+        );
+
+        $query = new Query($phpcrQuery, $this->eventDispatcher, $this->locale, [], $builder->getPrimaryAlias());
```
danrot, line 124: I don't like the fact that we have to pass the `EventDispatcher` here... But that's out of scope of this PR, and probably not possible in a different way with the current architecture.

lib/Subscriber/Phpcr/QuerySubscriber.php
```diff
@@ -39,13 +39,31 @@ class QuerySubscriber implements EventSubscriberInterface
...
      * @param EventDispatcherInterface $eventDispatcher
+     * @param QueryBuilderConverter $converter
+     * @param string $queryBuilderClass
      */
-    public function __construct(SessionInterface $session, EventDispatcherInterface $eventDispatcher)
+    public function __construct(
+        SessionInterface $session,
+        EventDispatcherInterface $eventDispatcher,
+        QueryBuilderConverter $builderConverter,
+        $queryBuilderClass = 'Doctrine\ODM\PHPCR\Query\Builder\QueryBuilder')
```
danrot: The bracket belongs to the next line together with the curling bracket.

danrot:
Have probably to look a bit deeper at it when I test it, otherwise it is hard to know what's really happening here...

lib/Query/QueryBuilderConverter.php
dantleech, line 218: Not exactly. They should do.

dantleech, line 218: I already do this elsewhere, if you want to change it it should be done globally.

dantleech:
U[pdated.

lib/Query/QueryBuilderConverter.php
danrot, line 218: That's the first time I see that, are you sure you've already used that in Sulu, and not somewhere else? I'd prefer to have something like this:

``` php
/**
 * @return string[] Some description explaining the two elements
 */
```

I am afraid that this style won't work on phpdocumentor, or that the syntax will look different, if something like this is supported later.

lib/Query/QueryBuilderConverter.php
```diff
@@ -0,0 +1,251 @@
...
+ */
+class QueryBuilderConverter extends ConverterPhpcr
+{
+    /**
+     * @var EventDispatcherInterface
+     */
+    private $eventDispatcher;
+
+    /**
+     * @var Metadata[]
```
danrot: This class is not used.

lib/Query/QueryBuilderConverter.php
```diff
@@ -0,0 +1,251 @@
...
+    public function __construct(
+        SessionInterface $session,
+        EventDispatcherInterface $eventDispatcher,
+        MetadataFactoryInterface $metadataFactory,
+        PropertyEncoder $encoder,
+        DocumentStrategyInterface $strategy
+    ) {
+        $this->eventDispatcher = $eventDispatcher;
+        $this->qomf = $session->getWorkspace()->getQueryManager()->getQOMFactory();
+        $this->metadataFactory = $metadataFactory;
```
danrot: Declaration missing

lib/Query/QueryBuilderConverter.php
```diff
@@ -0,0 +1,251 @@
...
+     * @param EventDispatcherInterface $eventDispatcher
+     * @param MetadataFactoryInterface $metadataFactory
+     * @param PropertyEncoder $encoder
+     */
+    public function __construct(
+        SessionInterface $session,
+        EventDispatcherInterface $eventDispatcher,
+        MetadataFactoryInterface $metadataFactory,
+        PropertyEncoder $encoder,
+        DocumentStrategyInterface $strategy
```
danrot: Missing in PHPDoc.

lib/Query/QueryBuilderConverter.php
```diff
@@ -0,0 +1,251 @@
...
+        if (!$this->locale) {
+            throw new \InvalidArgumentException(sprintf(
+                'No locale specified'
+            ));
+        }
+
+        $from = $builder->getChildrenOfType(QBConstants::NT_FROM);
+
+        if (!$from) {
+            throw new RuntimeException(
```
danrot: Usage missing.

lib/Query/QueryBuilderConverter.php
```diff
@@ -0,0 +1,251 @@
...
+            $query->setMaxResults($maxResults);
+        }
+
+        return $query;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    protected function applySourceConstraints(QueryBuilder $builder)
```
danrot: Parameter is not used?

lib/DocumentStrategyInterface.php
```diff
@@ -48,4 +50,48 @@ public function createNodeForDocument($document, NodeInterface $parentNode, $nam
...
+     * ```
+     *
+     * Can return NULL if no constraints should be added. This may be required
+     * if the primary type already represents the document class.
+     *
+     * @param QueryObjectModelFactoryInterface $qomf
+     * @param string $sourceAlias
+     * @param string $classFqn
+     *
+     * @return PHPCR\Query\QOM\ConstraintInterface
```
danrot: Leading `\` missing (the IDE doesn't recognize the type correctly otherwise)

lib/Metadata.php
```diff
@@ -82,6 +82,23 @@ public function getFieldMappings()
...
+            ));
+        }
+
+        return $this->fieldMappings[$name];
+    }
+
+    /**
      * @return string
      */
     public function getClass()
```
danrot: Not related to this PR, but this class contains a wrong phpdoc containing `ReflectionClass` instead of `\ReflectionClass`.

lib/Query/QueryBuilderConverter.php
danrot: It also looks like this method is not inherited, so the `{@inheritdoc}` doesn't make any sense.

lib/Subscriber/Phpcr/QuerySubscriber.php
```diff
@@ -39,13 +39,31 @@ class QuerySubscriber implements EventSubscriberInterface
...
+
+    /**
+     * @var string
+     */
+    private $queryBuilderClass;
+
+    /**
      * @param SessionInterface $session
      * @param EventDispatcherInterface $eventDispatcher
+     * @param QueryBuilderConverter $converter
```
danrot: Name does not match actual signature

tests/Functional/DocumentManager/QueryBuilderTest.php
```diff
@@ -0,0 +1,272 @@
...
+
+        $manager->flush();
+    }
+
+    /**
+     * It should query from an alias source.
+     */
+    public function testFromAlias()
+    {
+        $manager = $this->getDocumentManager();
```
danrot: The called method here is missing a correct `@return` statement.

tests/Functional/DocumentManager/QueryBuilderTest.php
```diff
@@ -0,0 +1,272 @@
...
+        $results = $builder->getQuery()->execute();
+
+        $this->assertEquals(1, count($results));
+        $this->assertContainsOnlyInstancesOf(FullDocument::class, $results->toArray());
+    }
+
+    /**
+     * It should throw an exception if the primary alias is not given when selecting from a join.
+     *
+     * @expectedException InvalidArgumentException
```
danrot: Leading `\` missing

tests/Functional/DocumentManager/QueryBuilderTest.php
```diff
@@ -0,0 +1,272 @@
...
+        $builder->setFirstResult(1);
+        $results = $builder->getQuery()->execute();
+        $this->assertCount(1, $results->toArray());
+    }
+
+    /**
+     * It should throw an exception if you try and select from multiple sources without
+     * specifying a primary selector.
+     *
+     * @expectedException InvalidArgumentException
```
danrot: Leading `\` missing

tests/Unit/MetadataTest.php
```diff
@@ -49,4 +49,28 @@ public function testReflectionClass()
...
+                'property' => 'foo',
+            ],
+            $this->metadata->getFieldMapping('hello')
+        );
+    }
+
+    /**
+     * It should throw an exception if the given mapping does not exist.
+     *
+     * @expectedException InvalidArgumentException
```
danrot: Leading `\` missing

tests/Unit/MetadataTest.php
```diff
@@ -49,4 +49,28 @@ public function testReflectionClass()
...
+     */
+    public function testGetFieldMappingNonExisting()
+    {
+        $this->metadata->getFieldMapping('hello');
+    }
 }
```
danrot: The variables in this class are missing PHPDoc

tests/Unit/Strategy/MixinStrategyTest.php
```diff
@@ -75,4 +80,38 @@ public function testResolveMetadataNotManaged()
 
         $this->assertNull($result);
     }
+
```
danrot: This class is missing all declarations.

tests/Unit/Subscriber/Phpcr/QuerySubscriberTest.php
```diff
@@ -86,10 +87,12 @@ public function setUp()
         $this->queryCreateEvent = $this->prophesize(QueryCreateEvent::class);
         $this->queryExecuteEvent = $this->prophesize(QueryExecuteEvent::class);
         $this->query = $this->prophesize(Query::class);
+        $this->converter = $this->prophesize(QueryBuilderConverter::class);
```
danrot: Declaration missing

lib/Query/QueryBuilderConverter.php
dantleech, line 218: fyi this syntax IS used for `@params` in PSR-5, however they havn't extended it to `@return` for some reason: https://github.com/phpDocumentor/fig-standards/blob/master/proposed/phpdoc.md#examples-12

dantleech:
Updated.

tests/Functional/DocumentManager/QueryBuilderTest.php
```diff
@@ -0,0 +1,272 @@
...
+    }
+
+    /**
+     * It should join sources, hydrating objects from right().
+     */
+    public function testQueryBuilderFromJoinHydrateRight()
+    {
+        $builder = $this->getDocumentManager()->createQueryBuilder();
+        $builder->setLocale('en');
+        $builder->from('i')->joinInner()
```
webmozart, line 167: Can the query builder API be simplified to be more like Doctrine's query builder API?

dantleech, line 167: Basically no ..

The PHPCR-ODM initially used the Doctrine API, but it made it impossible
to express some of semantics of PHPCR.

I wrote [this document](https://gist.github.com/dantleech/298c9618b5889bddeb87)
before writing the PHPCR-ODM query builder which explains in more depth
the problems that were faced.

webmozart, line 167: Thanks for the link. What about collapsing things like `->left()->document(x)` into `->left(x)`? Why `equi()` in one place and `eq()` in another?

dantleech, line 167: Because `document()` is only one type of "source". The source may also be another join.

The `equi` terminology stems from PHPCR (i.e. [JCR](https://www.day.com/maven/jsr170/javadocs/jcr-2.0/javax/jcr/query/qom/EquiJoinCondition.html) and means `equi join condition`. So a join based on an equality condition.

I agree that `->eq()` would be better than `->equi()`, the ODM query builder tried to copy as far as possible the PHPCR API.

But unless we intend to write a new query builder, this is out of scope, as these are all details of the PHPCR-ODM query builder.

PR #64, opened 2016-02-16
Added "isRegistered" method to inspector.

dantleech:
Allow the inspector to say if the document is currently known by the
system.

### DIFF
--- a/lib/DocumentInspector.php
+++ b/lib/DocumentInspector.php
@@ -191,4 +191,17 @@ class DocumentInspector
             ->getNodeForDocument($document)
             ->getIdentifier();
     }
+
+    /**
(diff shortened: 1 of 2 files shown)


danrot:
Why do we need that call on the `DocumentInspector`? The method looks to me like it is only acting as a proxy, why can't we ask the `DocumentRegistry` directly?

dantleech:
Basically beacause you don't want to inject a core and sensative part of
the document manager (the registry) into things just to be able to
determine if it is registered or not. The same goes for various other
methods on the inspector, ->getLocale, etc.

Saying that, depending on which direction we take re. the latest
refactorings it may no longer be necessary.

PR #66, opened 2016-02-23
Document context

dantleech:
fThis is the prototype for allowing multiple document manager instances.

This will allow consumers of this library to implement a document registry and retrieve different instances of the DocumentManager and each instance could be configured to use a different PHPCR repository.

In the context of Sulu this will be used to allow publishing from a `master` document manger to a `live` document manager which will be used when rendering the website.

Document managers now act as containers for their own services, and the document manager can be used to retrieve the services which are unique to it.

Events that a docuemnt manager emits now also include the document manager that emitted them, allowing event subscribers to act upon the correct document manager.
## Pros
- No duplication of event subscriber instances.
- Each DocumentManager encapsulates all of its dependencies.
- Significantly reduces the number of dependencies that event subscribers have.
## Cons
- 1 degree of separation: `$event->getManager()->getNodeManager()->find()`.
- DocumentManager instantiates some of its own dependencies.
- We currently need to expose "dependenceis" of the document manager to other services through the API -- the documetn manager should not allow this. But this may be better a case for refactoring the client services.


webmozart:
What is the issue that is being solved here?

dantleech:
Updated the description.

webmozart:
Thanks :)

dantleech:
Current benchmarks, this PR is slightly slower:

| its | subject | vcs_branch:develop:mode | vcs_branch:document_context:mode |
| --- | --- | --- | --- |
| 30 | benchHydrateMapping10 | 7.369ms | 7.731ms |
| 30 | benchHydrateMapping5 | 3.671ms | 3.882ms |

| its | subject | vcs_branch:develop:mode | vcs_branch:document_context:mode |
| --- | --- | --- | --- |
| 30 | benchPersistMapping5 | 1.979ms | 2.070ms |
| 30 | benchPersistMapping10 | 2.952ms | 3.004ms |

| its | subject | vcs_branch:develop:mode | vcs_branch:document_context:mode |
| --- | --- | --- | --- |
| 30 | benchBuild | 0.014ms | 0.014ms |

| its | subject | vcs_branch:develop:mode | vcs_branch:document_context:mode |
| --- | --- | --- | --- |
| 30 | benchCreatePersist | 528.830ms | 551.716ms |
| 30 | benchCreatePersistPhpcr | 303.928ms | 305.534ms |

lib/Collection/ChildrenCollection.php
```diff
@@ -75,7 +69,8 @@ public function current()
         $childNode = $this->documents->current();
 
         $hydrateEvent = new HydrateEvent($childNode, $this->locale, $this->options);
-        $this->dispatcher->dispatch(Events::HYDRATE, $hydrateEvent);
+        $hydrateEvent->attachContext($this->context);
```
dantleech: `attachContext`: We either inject the context in the constructor, or "attach it". Attaching allows us to avoid injecting into the constructor, which is basically annoying as all events need to take this into account when overriding the `__constructor`.

It may indeed be better to inject into the constructor however..

lib/DocumentManager.php
```diff
@@ -130,8 +180,8 @@ public function flush()
      */
     public function clear()
     {
-        $event = new Event\ClearEvent();
-        $this->eventDispatcher->dispatch(Events::CLEAR, $event);
+        $event = new Event\ClearEvent($this->context);
```
dantleech: This argument does not exist .. same with `flush`

lib/DocumentManager.php
```diff
@@ -150,13 +200,37 @@ public function createQuery($query, $locale = null, array $options = [])
...
+    {
+        return $this->nodeManager;
+    }
+
+    public function getRegistry()
+    {
+        return $this->registry;
+    }
+
+    public function getProxyFactory()
```
dantleech: These methods (`getNodeManager`, `getRegistry` and `getProxyFactory`) are required by the legacy Sulu system which is not using events. The API should NOT be exposing these methods, but not currently sure of a better way to handle this.

lib/DocumentManager.php
```diff
@@ -169,10 +243,16 @@ private function getOptionsResolver($eventName)
...
         $event = new Event\ConfigureOptionsEvent($resolver);
-        $this->eventDispatcher->dispatch(Events::CONFIGURE_OPTIONS, $event);
+        $this->dispatch(Events::CONFIGURE_OPTIONS, $event);
 
         $this->optionsResolvers[$eventName] = $resolver;
 
         return $resolver;
     }
+
+    private function dispatch($eventName, AbstractEvent $event)
```
dantleech: Doc

lib/Event/AbstractEvent.php
```diff
@@ -11,10 +11,29 @@
...
 abstract class AbstractEvent extends Event
 {
+    private $context;
+
+    public function attachContext(DocumentManagerContext $context)
+    {
+        $this->context = $context;
+    }
+
+    public function getContext()
```
dantleech: Doc

lib/ProxyFactory.php
```diff
@@ -30,38 +30,33 @@ class ProxyFactory
...
-        DocumentRegistry $registry,
         MetadataFactoryInterface $metadataFactory
     ) {
         $this->proxyFactory = $proxyFactory;
-        $this->dispatcher = $dispatcher;
-        $this->registry = $registry;
         $this->metadataFactory = $metadataFactory;
     }
 
+    public function attachContext(DocumentManagerContext $context)
```
dantleech: Add interface if kept.

lib/Strategy/MixinStrategy.php
```diff
@@ -43,7 +44,12 @@ public function createNodeForDocument($document, NodeInterface $parentNode, $nam
 
         $node = $parentNode->addNode($name);
         $node->addMixin($metadata->getPhpcrType());
-        $node->setProperty('jcr:uuid', UUIDHelper::generateUUID());
+
+        if (!$document instanceof UuidBehavior || !$uuid = $document->getUuid()) {
+            $uuid = UUIDHelper::generateUUID();
```
dantleech, line 56: Add comment about what this does: it will keep existing UUIDs set on new documents (which will happen when persisting a document from one DM to another).

lib/DocumentInspectorFactory.php
```diff
@@ -0,0 +1,48 @@
...
+     * @var DocumentInspector
+     */
+    private $inspector;
+
+    public function __construct(PathSegmentRegistry $pathSegmentRegistry)
+    {
+        $this->pathSegmentRegistry = $pathSegmentRegistry;
+    }
+
+    public function getInspector(DocumentManagerContext $context)
```
dantleech, line 37: inheritdoc

lib/DocumentManager.php
```diff
@@ -27,11 +30,58 @@ class DocumentManager implements DocumentManagerInterface
...
+     */
+    private $proxyFactory;
+
+    /**
      * @param EventDispatcherInterface $eventDispatcher
      */
-    public function __construct(EventDispatcherInterface $eventDispatcher)
-    {
+    public function __construct(
+        SessionInterface $session,
```
dantleech: rm doc

lib/ProxyFactory.php
dantleech: It is only used now for this factory.

lib/Event/AbstractDocumentEvent.php
```diff
@@ -21,8 +23,9 @@
     /**
      * @param object $document
      */
-    public function __construct($document)
+    public function __construct(DocumentManagerContext $context, $document)
```
dantleech, line 27: doc

lib/Event/CreateEvent.php
```diff
@@ -26,8 +28,9 @@ class CreateEvent extends AbstractEvent
     /**
      * @param string $alias
      */
-    public function __construct($alias)
+    public function __construct(DocumentManagerContext $context, $alias)
```
dantleech, line 33: Doc

lib/Event/FindEvent.php
```diff
@@ -37,8 +38,9 @@ class FindEvent extends AbstractEvent
      * @param string $locale
      * @param array $options
      */
-    public function __construct($identifier, $locale, array $options = [])
+    public function __construct(DocumentManagerContext $context, $identifier, $locale, array $options = [])
```
dantleech: Dc

lib/Event/HydrateEvent.php
```diff
@@ -20,8 +21,9 @@ class HydrateEvent extends AbstractMappingEvent
      * @param string $locale
      * @param array $options
      */
-    public function __construct(NodeInterface $node, $locale, array $options = [])
+    public function __construct(DocumentManagerContext $context, NodeInterface $node, $locale, array $options = [])
```
dantleech: Fc

lib/Event/MoveEvent.php
```diff
@@ -32,8 +34,9 @@ class MoveEvent extends AbstractEvent
      * @param object $document
      * @param string $destId
      */
-    public function __construct($document, $destId)
+    public function __construct(DocumentManagerContext $context, $document, $destId)
```
dantleech: Dc

lib/Event/QueryCreateEvent.php
```diff
@@ -44,8 +45,9 @@ class QueryCreateEvent extends AbstractEvent
      * @param array $options
      * @param null|string $primarySelector
      */
-    public function __construct($innerQuery, $locale, array $options = [], $primarySelector = null)
+    public function __construct(DocumentManagerContext $context, $innerQuery, $locale, array $options = [], $primarySelector = null)
```
dantleech: Dc + all docs for all events.

lib/ProxyFactory.php
dantleech: Dc

tests/Bootstrap.php
```diff
@@ -144,9 +145,16 @@ public static function createDbalConnection()
             'host' => 'localhost',
             'user' => 'admin',
             'password' => 'admin',
-            'path' => __DIR__ . '/../data/test.sqlite',
+            //'path' => __DIR__ . '/../data/test.sqlite',
```
dantleech: rm

tests/Unit/DocumentInspectorFactoryTest.php
```diff
@@ -0,0 +1,63 @@
...
+    private $proxyFactory;
+
+    public function setUp()
+    {
+        $this->pathSegmentRegistry = $this->prophesize(PathSegmentRegistry::class);
+        $this->registry = $this->prophesize(DocumentRegistry::class);
+        $this->proxyFactory = $this->prophesize(ProxyFactory::class);
+        $this->context = $this->prophesize(DocumentManagerContext::class);
+        $this->context->getRegistry()->willReturn($this->registry->reveal());
+        $this->context->getProxyFactory()->willReturn($this->proxyFactory->reveal());
```
dantleech: contrent property.b

lib/Event/ConfigureOptionsEvent.php
```diff
@@ -12,8 +12,10 @@
 namespace Sulu\Component\DocumentManager\Event;
 
 use Symfony\Component\OptionsResolver\OptionsResolver;
+use Sulu\Component\DocumentManager\DocumentManagerContext;
+use Symfony\Component\EventDispatcher\Event;
 
-class ConfigureOptionsEvent extends AbstractEvent
+class ConfigureOptionsEvent extends Event
```
dantleech: This will break the Debug event dispatcher. Either all events need to extend AbstractEvent (or DebugEvent) or the DebugEventDispatcher needs to handle non-AbstractEvent classes.

lib/Metadata.php
```diff
@@ -149,4 +149,19 @@ public function setPhpcrType($phpcrType)
     {
         $this->phpcrType = $phpcrType;
     }
+
+    public function setFieldValue($document, $field, $value)
```
dantleech, line 166: Doc and test

lib/Metadata.php
```diff
@@ -149,4 +149,19 @@ public function setPhpcrType($phpcrType)
...
+                'Field "%s" is not mapped. Mapped fields: "%s"',
+                $field, implode('", "', array_keys($this->fieldMappings))
+            ));
+        }
+
+        $reflection = $this->getReflectionClass();
+        $property = $reflection->getProperty($field);
+        $property->setAccessible(true);
+        $property->setValue($document, $value);
+    }
```
dantleech: Add corresponding getFieldValue

lib/NodeManager.php
```diff
@@ -57,6 +58,10 @@ public function find($identifier)
             throw new DocumentNotFoundException(sprintf(
                 'Could not find document with ID or path "%s"', $identifier
             ), null, $e);
+        } catch (ItemNotFoundException $e) {
```
dantleech: Test + comment -- why do we recieve different exceptions when a path is not found?

lib/NodeManager.php
```diff
@@ -150,10 +154,11 @@ public function clear()
      * Create a path.
      *
      * @param string $path
+     * @param bool $generateUuid
      *
      * @return NodeInterface
      */
-    public function createPath($path)
+    public function createPath($path, $generateUuid = true)
```
dantleech: Test -- is this the best solution

lib/Strategy/MixinStrategy.php
```diff
@@ -43,7 +44,19 @@ public function createNodeForDocument($document, NodeInterface $parentNode, $nam
...
-        $node->setProperty('jcr:uuid', UUIDHelper::generateUUID());
+
+        // We only generate a new UUID if the document does not already have one.
+        //
+        // If the document DOES already have a UUID then that implies that it was previously
+        // managed by a different document manager and we assume that the UUID should be
+        // preserved (as in the case where we want to preserve the UUID between multiple workspaces).
+        //
+        // In the future this behavior could be made configurable if required.
+        if (!$document instanceof UuidBehavior || !$uuid = $document->getUuid()) {
```
dantleech: Test

dantleech:
Consider removing the DocumentManagreContext -- we already have had to add accessors to the DocumentManager - we may as well pass the DocumentManager itself in the event. This is also the approach the ORM takes.

lib/Subscriber/Behavior/Mapping/ParentSubscriber.php
```diff
@@ -90,6 +60,13 @@ public function handleSetParentNodeFromDocument(PersistEvent $event)
     {
         $document = $event->getDocument();
 
+        $options = $event->getOptions();
+
+        // path takes precedence over this subscriber.
+        if (isset($options['path']) && $options['path']) {
```
dantleech: Not sure if there is a better way to handle this. Add test.

lib/Subscriber/Behavior/Path/AbstractFilingSubscriber.php
```diff
@@ -53,9 +39,16 @@ public function handlePersist(PersistEvent $event)
             return;
         }
 
+        $options = $event->getOptions();
+
+        // if "path" option has been explicitly set.
+        if (isset($options['path']) || isset($options['parent_path'])) {
```
dantleech: TRest

lib/Event/AbstractManagerEvent.php
```diff
@@ -0,0 +1,52 @@
...
+    public function __construct(DocumentManagerInterface $manager)
+    {
+        $this->manager = $manager;
+    }
+
+    public function getManager()
+    {
+        if (null === $this->manager) {
+            throw new \RuntimeException(
+                'No DocumentManagerInterface has been set on this event, maybe this class has overridden the constructor and forgotten about it?'
```
dantleech: .. instance

lib/Collection/QueryResultCollection.php
```diff
@@ -52,21 +47,19 @@ class QueryResultCollection extends AbstractLazyCollection
...
-     * @param QueryResultInterface $result
-     * @param EventDispatcherInterface $eventDispatcher
-     * @param string $locale
-     * @param array $options
-     * @param null|string $primarySelector
+     * @var DocumentManagerInterface
      */
+    private $manager;
+
     public function __construct(
```
danrot, line 55: I think @chirimoya recently said that he'd like to have the phpdoc also for constructor calls.

lib/DocumentInspectorFactory.php
```diff
@@ -0,0 +1,51 @@
...
+            return $this->inspector;
+        }
+
+        $this->inspector = new DocumentInspector(
+            $manager->getRegistry(),
+            $this->pathSegmentRegistry,
+            $manager->getProxyFactory()
+        );
+
+        return $this->inspector;
```
danrot: From what I see here I can't recognize why a `DocumentInspectorFactory` is necessary. Once the value is cached always the same inspector is returned. Is it necessary to have one of them for each `DocumentManager`? Then they should also be cached separately, otherwise I would remove this class.

lib/DocumentManager.php
```diff
@@ -27,11 +39,62 @@ class DocumentManager implements DocumentManagerInterface
...
+        // document manager instances.
+        $this->metadataFactory = $metadataFactory;
+
+        // the PHPCR session SHOULD be unique to all document managers.
+        $this->session = $session;
+
+        // instantiate other objects scoped to this document manager.
+        $this->nodeManager = new NodeManager($session);
+        $this->proxyFactory = new ProxyFactory($this, $lazyProxyFactory, $metadataFactory);
+        $this->registry = new DocumentRegistry($defaultLocale);
```
danrot: Why not also do so with the `DocumentInspector`? Don't say it would be better, but at least more consistent. However I am not really sure about that at the moment.

lib/DocumentManager.php
```diff
@@ -130,7 +193,7 @@ public function flush()
      */
     public function clear()
     {
-        $event = new Event\ClearEvent();
+        $event = new Event\ClearEvent($this, $this);
```
danrot: Do we really have to pass `$this` twice?

lib/DocumentManagerInterface.php
```diff
@@ -122,4 +124,56 @@ public function createQuery($query, $locale = null, array $options = []);
      * http://doctrine-phpcr-odm.readthedocs.org/en/latest/reference/query-builder.html
      */
     public function createQueryBuilder();
+
+    /**
+     * Retrieve the document inspector.
+     *
+     * NOTE: In the future this will be a factory for retrieving inspectors
+     * for a given object/class.
```
danrot, line 130: Ok, that's why there is this one factory class at the moment... Don't like the idea, since the `DocumentInspector` is just a temporary hack IMO. The document should be able to tell stuff like that on its own, there is also no real equivalent to the `DocumentInspector` in the ORM.

lib/DocumentManagerInterface.php
```diff
@@ -122,4 +124,56 @@ public function createQuery($query, $locale = null, array $options = []);
...
+     * @return SessionInterface
+     */
+    public function getSession();
+
+    /**
+     * Return the proxy factory instance.
+     *
+     * @return ProxyFactory
+     */
+    public function getProxyFactory();
```
danrot: Is it really necessary to put all of these into this interface? Until now I thought as the `DocumentManager` as a facade for a more complex underlying system. This way we populate quite a lot of these complex classes, making later refactorings pretty hard.

lib/Event/AbstractManagerEvent.php
```diff
@@ -0,0 +1,52 @@
...
+
+use Sulu\Component\DocumentManager\DocumentManagerInterface;
+use Symfony\Component\EventDispatcher\Event;
+
+/**
+ * Abstract class for events which require the DocumentManagerInterface - i.e.
+ * all events for which subscribers require access to a document manager or any
+ * of its dependencies.
+ */
+abstract class AbstractManagerEvent extends AbstractEvent
```
danrot: Any real need for this class? Why not just combining it with the `AbstractDocumentEvent`? Or the `AbstractEvent`? This chain of inheritance gets quite long this way.

lib/Event/AbstractManagerEvent.php
```diff
@@ -0,0 +1,52 @@
...
+    public function __construct(DocumentManagerInterface $manager)
+    {
+        $this->manager = $manager;
+    }
+
+    public function getManager()
+    {
+        if (null === $this->manager) {
+            throw new \RuntimeException(
+                'No DocumentManagerInterface instance has been set on this event, maybe this class has overridden the constructor and forgotten about it?'
```
danrot: line too long?

danrot: And is this check necessary at all? The constructor always sets the manager, and because of the typehint it's not allowed to be null.

lib/NodeManager.php
```diff
@@ -57,6 +65,10 @@ public function find($identifier)
             throw new DocumentNotFoundException(sprintf(
                 'Could not find document with ID or path "%s"', $identifier
             ), null, $e);
+        } catch (ItemNotFoundException $e) {
+            throw new DocumentNotFoundException(sprintf(
+                'Could not find document with ID or path "%s"', $identifier
```
danrot: And we are already duplicating the exception message, because it's set in the wrong place 😜

lib/ProxyFactory.php
```diff
@@ -30,35 +30,27 @@ class ProxyFactory
...
+        DocumentManagerInterface $documentManager,
         LazyLoadingGhostFactory $proxyFactory,
-        EventDispatcherInterface $dispatcher,
-        DocumentRegistry $registry,
         MetadataFactoryInterface $metadataFactory
     ) {
+        $this->documentManager = $documentManager;
         $this->proxyFactory = $proxyFactory;
-        $this->dispatcher = $dispatcher;
-        $this->registry = $registry;
```
danrot, line 61: Why not keeping this variable here, but instantiate it with `$this->documentManager->getRegistry()`? Would result in less changes and methods calls in this file.

lib/Subscriber/Behavior/Path/AliasFilingSubscriber.php
```diff
@@ -77,12 +57,14 @@ protected function supports($document)
     }
 
     /**
-     * @param $document
+     * @param object $document
```
danrot: wrong comment

tests/Unit/Subscriber/Behavior/Audit/TimestampSubscriberTest.php
```diff
@@ -16,8 +16,9 @@
 use Sulu\Component\DocumentManager\DocumentAccessor;
 use Sulu\Component\DocumentManager\Event\PersistEvent;
 use Sulu\Component\DocumentManager\Subscriber\Behavior\Audit\TimestampSubscriber;
+use Sulu\Component\DocumentManager\Tests\Unit\Subscriber\Behavior\SubscriberTestCase;
 
-class TimestampSubscriberTest extends \PHPUnit_Framework_TestCase
+class TimestampSubscriberTest extends SubscriberTestCase
```
danrot: Why chnange to the `SubscriberTestCase` if it was obviously also working before?

tests/Unit/Subscriber/Behavior/SubscriberTestCase.php
```diff
@@ -0,0 +1,16 @@
...
+ *
+ * (c) MASSIVE ART WebServices GmbH
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace Sulu\Component\DocumentManager\Tests\Unit\Subscriber\Behavior;
+
+class SubscriberTestCase extends \PHPUnit_Framework_TestCase
```
danrot: What's that for?

lib/DocumentManagerInterface.php
dantleech, line 130: In the ORM the equivilent of the document instpector is the
EntityManager/DocumentManager, which is used to introspect the document.

The reason I introduced this class was to avoid API bloat for the
DocumentManager and to eventually allow a flexible way to introduce
introspection methods.

The problem with the Douments "knowing" about this stuff is that the
end-user needs to implement that stuff. If we want to know which
webspace a document belongs to we need to make `->getWebspaceName()` a
mandatory method on all documents, etc.

lib/DocumentManagerInterface.php
dantleech: This was a more difficult decision. All of these things are scoped to
the document manager instance. They belong to the document manager and
are not shared by anything else (with the exception of the
MetadataFactory, but in the future that might change).

The alternative was the `DocumentContext` which was an accessor for the
services contained within the DocumentManager. But in practice this
seemed like a pointless step.

The Doctrine Entity/Document Managers all provide access to their
various components through their Managers.

lib/Event/AbstractManagerEvent.php
dantleech: It is overridden by other classes which override the __constructor.
Better safe than sorry.

lib/ClassNameInflector.php
```diff
@@ -36,10 +36,27 @@ class ClassNameInflector
...
+    }
+
+    /**
+     * Return true if the given class name appears to be the name of a proxy class.
+     *
+     * @param string $className
+     *
+     * @return bool
+     */
+    public static function isProxyClassName($className)
```
dantleech: Assuming this is needed in Sulu

tests/Unit/Subscriber/Behavior/SubscriberTestCase.php
dantleech: An unfinished improvement I think. The subscriber tests share a lot of requirements, the idea is to reduce the boilerplate code.

I think this is a natural evolution, so despite the fact that there is nothing in this class we should keep it.

lib/ClassNameInflector.php
danrot: Are you assuming, or is it really needed? It was obviously not needed until now, so have you added calls to this method?

lib/Collection/ChildrenCollection.php
```diff
@@ -49,19 +45,18 @@ class ChildrenCollection extends AbstractLazyCollection
     private $initialized = false;
 
     /**
-     * @param NodeInterface $parentNode
-     * @param EventDispatcherInterface $dispatcher
-     * @param string $locale
-     * @param array $options
+     * @var DocumentManagerInterface
```
danrot: `DocumentManagerContext`

lib/DocumentAccessor.php
```diff
@@ -13,6 +13,7 @@
 
 use ProxyManager\Proxy\LazyLoadingInterface;
 use Sulu\Component\DocumentManager\Exception\DocumentManagerException;
```
danrot, line 15: Only here because of wrong comment.

lib/DocumentInspectorFactory.php
```diff
@@ -0,0 +1,52 @@
...
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace Sulu\Component\DocumentManager;
+
+/**
+ * Document inspector factory for the default DocumentInspector.
+ */
+class DocumentInspectorFactory implements DocumentInspectorFactoryInterface
```
danrot, line 17: Didn't we say we remove this class? I thought you've added the comment somewhere, but I can't find it anymore... Shouldn't that have been merged with the contexts (because I've already seen the contexts in here)?

lib/DocumentManager.php
```diff
@@ -11,9 +11,26 @@
...
+ *   within the scope of the particular document manager (as there can be many).
+ *
+ * This class is therefore an event dispatcher and a service container and has
+ * no other business logic within it.
+ *
+ * All primary methods are wrapped in a try catch block, the exception is then
+ * wrapped in a `DocumentManagerException` with a reference to this document
+ * managers name.
+ *
+ * TODO: Create a proxy document manager which automatically wraps all calls.
```
danrot, line 32: What would you need that for?

lib/DocumentManager.php
```diff
@@ -22,138 +39,211 @@ class DocumentManager implements DocumentManagerInterface
...
+            $options = $this->getOptionsResolver(Events::FIND)->resolve($options);
 
-        $event = new Event\FindEvent($identifier, $locale, $options);
-        $this->eventDispatcher->dispatch(Events::FIND, $event);
+            $event = new Event\FindEvent($this->context, $identifier, $locale, $options);
+            $this->eventDispatcher->dispatch(Events::FIND, $event);
 
-        return $event->getDocument();
+            return $event->getDocument();
+        } catch (\Exception $e) {
```
danrot, line 81: Didn't we finally agree that catching all exceptions isn't a good idea?

lib/DocumentManager.php
```diff
@@ -175,4 +265,34 @@ private function getOptionsResolver($eventName)
...
     }
+
+    /**
+     * If the exception is an instanceof `DocumentManagerException` set the
+     * document manager name that threw.
+     *
+     * @param \Exception $e
+     * @param string $context
+     */
+    private function processException(\Exception $exception, $context)
```
danrot, line 276: Why not accepting only `DocumentManagerException`s, only catch `DocumentManagerException`s above, and remove the if in this method?

lib/DocumentManagerContext.php
```diff
@@ -0,0 +1,199 @@
...
+
+        // the PHPCR session SHOULD be unique to all (end user) document
+        // manager contexts.
+        $this->session = $session;
+
+        // instantiate other objects scoped to this document manager.
+        $this->nodeManager = new NodeManager($session);
+        $this->proxyFactory = new ProxyFactory($this, $lazyProxyFactory, $metadataFactory);
+        $this->registry = new DocumentRegistry($defaultLocale);
+        $this->manager = new DocumentManager($this);
```
danrot, line 101: I don't really get the separation of creating new instances vs. injecting something. I would simply inject all the services, so that they are also available from the container if necessary. Or if we are consequent the other way round we should also create the `EventDispatcher` like that. But I would like to decide that together (or didn't we already say that we export that kind of stuff to the Extension?)

lib/DocumentManagerInterface.php
danrot, line 130: Just thinking: Why not getting the inspector from the context?

lib/DocumentRegistry.php
```diff
@@ -261,7 +262,7 @@ private function assertDocumentExists($document)
         $oid = spl_object_hash($document);
 
         if (!isset($this->documentMap[$oid])) {
-            throw new \RuntimeException(sprintf(
+            throw new RuntimeException(spl_object_hash($this) . sprintf(
```
danrot: Does this `spl_object_hash` make sense here?

lib/Event/CreateEvent.php
```diff
@@ -24,10 +27,12 @@ class CreateEvent extends AbstractEvent
     private $alias;
 
     /**
+     * @param DocumentManagerContext $manager
      * @param string $alias
      */
-    public function __construct($alias)
+    public function __construct(DocumentManagerContext $manager, $alias)
```
danrot: `$context`

lib/Event/FindEvent.php
danrot: What do you mean by that?

lib/EventDispatcher/DebugEventDispatcher.php
```diff
@@ -23,6 +23,11 @@
 class DebugEventDispatcher extends ContainerAwareEventDispatcher
 {
     /**
+     * @var int
+     */
+    private $depth = 0;
```
danrot, line 28: Is it safe to have that as a member variable? Doesn't feel too good...

lib/DocumentInspectorFactory.php
dantleech, line 17: We did, but at this stage it requires more effort than it is worth, and
then I am not sure if it is the best thing to do. Would rather wait
before removing this, if at all.

lib/DocumentManager.php
dantleech, line 32: Because otherwise we have a try/catch block for each method call, when
we should be saying that ALL method calls should go through a certain
gateway which is wrapped in a single try/catch block.

lib/Exception/DocumentManagerException.php
```diff
@@ -11,6 +11,59 @@
...
+            $this->formatMessage($documentManagerName);
+        }
+    }
+
+    /**
+     * Set the name of the document manager which threw this exception.
+     *
+     * @param string $name
+     */
+    public function setDocumentManagerName($name)
```
danrot, line 40: Why do we need a setter here? This is quite confusing...

lib/DocumentManagerInterface.php
dantleech, line 130: This is the Context .. ?

lib/DocumentRegistry.php
dantleech: No this was debugging.

lib/Event/FindEvent.php
dantleech: Doc

lib/EventDispatcher/DebugEventDispatcher.php
dantleech, line 28: I don't see why not. It tracks the current nesting level.

lib/Exception/DocumentManagerException.php
dantleech, line 40: Because the thing that threw this exception probably does not know which document manager invoked it. This is why we catch the Exceptions in the DocumentManager - to set the name.

lib/ProxyFactory.php
```diff
@@ -30,35 +30,27 @@ class ProxyFactory
...
     /**
-     * @var EventDispatcherInterface
+     * @var MetadataFactoryInterface
      */
-    private $dispatcher;
+    private $metadataFactory;
 
     /**
-     * @var DocumentRegistry
+     * @var DocumentManagerInterface
```
danrot: context

lib/Subscriber/Behavior/Path/AutoNameSubscriber.php
```diff
@@ -205,8 +192,9 @@ private function handleMoveCopy(MoveEvent $event)
         }
 
         $destId = $event->getDestId();
-        $node = $this->registry->getNodeForDocument($document);
-        $destNode = $this->nodeManager->find($destId);
+        $manager = $event->getManager();
```
danrot: Doesn't seem to be used.

lib/NodeManager.php
dantleech: Not sure I understand, but we need not duplicate the message no.

lib/ClassNameInflector.php
dantleech: Need to check.

lib/Subscriber/Behavior/Path/ExplicitSubscriber.php
```diff
@@ -141,36 +138,74 @@ public function handlePersist(PersistEvent $event)
...
+     * @param NodeInterface $node  The existing node.
+     * @param NodeInterface $parentNode  The parent node from the event (not necessarily the subject node)
+     * @param string $nodeName  The target name of the node.
+     */
+    private function handleExisting(NodeManager $nodeManager, NodeInterface $node, NodeInterface $parentNode, $nodeName)
     {
+        $realParentPath = PathHelper::getParentPath($node->getPath());
+        $eventParentPath = $parentNode->getPath();
+
+        if ($eventParentPath !== $realParentPath) {
```
danrot, line 163: Shouldn't changing of the parent path be handled by the `ParentSubscriber`?

lib/DocumentManager.php
dantleech, line 276: Less flexibility.

lib/Subscriber/Core/MappingSubscriber.php
```diff
@@ -220,6 +210,8 @@ public function handleHydrate(AbstractMappingEvent $event)
         $node = $event->getNode();
         $accessor = $event->getAccessor();
         $document = $event->getDocument();
+        $manager = $event->getManager();
```
danrot: Doesn't seem to be used.

lib/Subscriber/Core/MappingSubscriber.php
```diff
@@ -274,11 +268,19 @@ private function hydrateReferenceField(
...
-                $accessor->set(
+            try {
+                if ($referencedNode) {
+                    $accessor->set(
                     $fieldName,
-                    $this->proxyFactory->createProxyForNode($document, $referencedNode, $options)
+                    $proxyFactory->createProxyForNode($document, $referencedNode, $options)
                 );
+                }
+            } catch (\Exception $e) {
```
danrot, line 278: Isn't there a more specific exception you could catch?

lib/Subscriber/Phpcr/ReorderSubscriber.php
```diff
@@ -68,28 +47,29 @@ public function handleReorder(ReorderEvent $event)
         $document = $event->getDocument();
         $siblingId = $event->getDestId();
         $after = $event->getAfter();
+        $manager = $event->getManager();
```
danrot: Doesn't seem to be used.

lib/DocumentManagerContext.php
dantleech, line 101: For example, as you suggest, we would here inject the PHPCR session and the NodeManager (which contains the PHPCR session) `__construct($phpcrSession, $nodeManager, ..)`. 

It would make no sense for the NodeManager to have a Session which is different from the PhpcrSession we provide to the context.

Otherwise: 
1. it provides a clear entry point for the DocumentManager. To create a document manager instance, you just need to instantiate the Context, you do not need to instantiate several interdependent object instances - and potentially fuck it up.
2. The DocumentManager requires an instance of the `Context`, which requires an instance of the `DocumentManager`. Making the `Context` a factory for the `DocumentManager` is better than playing some strange `chicken and egg` game (the reason for the previous `Context::attachDocumentManager` method in previous PR version).

I have considered, and tried, instantaiting everything, but it really doesn't make sense to me for the above reasons.

tests/Bench/BaseBench.php
```diff
@@ -50,7 +50,7 @@ protected function getSession()
         return $session;
     }
 
-    protected function getDocumentManager()
+    protected function getManager()
```
danrot, line 53: I wouldn't rename that method, makes it less expressive.

tests/Functional/BaseTestCase.php
```diff
@@ -46,7 +46,7 @@ protected function getContainer()
         return $this->container;
     }
 
-    protected function getDocumentManager()
+    protected function getManager()
```
danrot, line 49: Same as above, would not rename that method.

tests/Unit/Collection/ChildrenCollectionTest.php
```diff
@@ -25,10 +27,13 @@ public function setUp()
         $this->parentNode = $this->prophesize(NodeInterface::class);
 
         $this->dispatcher = $this->prophesize(EventDispatcherInterface::class);
+        $this->manager = $this->prophesize(DocumentManagerInterface::class);
```
danrot: Doesn't seem to be used.

tests/Unit/Collection/QueryResultCollectionTest.php
```diff
@@ -25,10 +27,13 @@ public function setUp()
     {
         $this->queryResult = $this->prophesize(QueryResultInterface::class);
         $this->dispatcher = $this->prophesize(EventDispatcherInterface::class);
+        $this->manager = $this->prophesize(DocumentManagerInterface::class);
```
danrot: Doesn't seem to be used.

tests/Unit/Collection/ChildrenCollectionTest.php
```diff
@@ -25,10 +27,13 @@ public function setUp()
...
+        $this->context = $this->prophesize(DocumentManagerContext::class);
```
danrot, line 54: File is lacking all member variable declarations.

tests/Unit/Collection/QueryResultCollectionTest.php
```diff
@@ -25,10 +27,13 @@ public function setUp()
...
+        $this->context = $this->prophesize(DocumentManagerContext::class);
+        $this->context->getEventDispatcher()->willReturn($this->dispatcher->reveal());
```
danrot, line 50: File is lacking all member variable declarations.

tests/Unit/Collection/ReferrerCollectionTest.php
```diff
@@ -27,10 +29,13 @@ public function setUp()
         $this->node = $this->prophesize(NodeInterface::class);
 
         $this->dispatcher = $this->prophesize(EventDispatcherInterface::class);
+        $this->manager = $this->prophesize(DocumentManagerInterface::class);
+        $this->context = $this->prophesize(DocumentManagerContext::class);
+        $this->context->getEventDispatcher()->willReturn($this->dispatcher->reveal());
```
danrot, line 62: File is lacking all member variable declarations.

tests/Unit/Collection/ReferrerCollectionTest.php
```diff
@@ -27,10 +29,13 @@ public function setUp()
         $this->node = $this->prophesize(NodeInterface::class);
 
         $this->dispatcher = $this->prophesize(EventDispatcherInterface::class);
+        $this->manager = $this->prophesize(DocumentManagerInterface::class);
```
danrot: Doesn't seem to be used.

lib/Subscriber/Behavior/Path/ExplicitSubscriber.php
dantleech, line 163: I actually have no idea why this method was added.

lib/Subscriber/Core/MappingSubscriber.php
dantleech, line 278: By catching the all exceptions we get to see exactly `which field` has the error - which is a crucial piece of information, if we do not catch, then we lose this information.

tests/Unit/Subscriber/Behavior/Mapping/LocaleSubscriberTest.php
```diff
@@ -22,15 +23,18 @@ class LocaleSubscriberTest extends \PHPUnit_Framework_TestCase
...
         $this->accessor = new DocumentAccessor($this->document);
         $this->registry = $this->prophesize(DocumentRegistry::class);
+        $this->manager = $this->prophesize(DocumentManagerInterface::class);
+        $this->hydrateEvent = $this->prophesize(HydrateEvent::class);
+        $this->hydrateEvent->getRegistry()->willReturn($this->registry->reveal());
+        $this->hydrateEvent->getManager()->willReturn($this->manager->reveal());
 
         $this->subscriber = new LocaleSubscriber(
-            $this->registry->reveal()
+            $this->manager->reveal()
```
danrot: The subscriber doesn't take any arguments anymore.

tests/Unit/Subscriber/Behavior/Mapping/LocaleSubscriberTest.php
```diff
@@ -22,15 +23,18 @@ class LocaleSubscriberTest extends \PHPUnit_Framework_TestCase
 {
     public function setUp()
     {
-        $this->hydrateEvent = $this->prophesize(HydrateEvent::class);
         $this->notImplementing = new \stdClass();
         $this->node = $this->prophesize(NodeInterface::class);
         $this->document = new TestLocaleDocument();
         $this->accessor = new DocumentAccessor($this->document);
         $this->registry = $this->prophesize(DocumentRegistry::class);
+        $this->manager = $this->prophesize(DocumentManagerInterface::class);
```
danrot: When removed from the constructor below this can also be removed. And also this file is missing all member variable declarations.

tests/Unit/Subscriber/Behavior/Mapping/ParentSubscriberTest.php
```diff
@@ -81,6 +84,7 @@ class ParentSubscriberTest extends \PHPUnit_Framework_TestCase
     public function setUp()
     {
         $this->hydrateEvent = $this->prophesize(HydrateEvent::class);
+        $this->persistEvent = $this->prophesize(PersistEvent::class);
```
danrot, line 92: Missing this and some other member variable declarations.

tests/Unit/Subscriber/Behavior/Mapping/PathSubscriberTest.php
```diff
@@ -31,9 +32,12 @@ public function setUp()
         $this->inspector = $this->prophesize(DocumentInspector::class);
         $this->accessor = $this->prophesize(DocumentAccessor::class);
         $this->hydrateEvent->getAccessor()->willReturn($this->accessor);
+        $this->manager = $this->prophesize(DocumentManagerInterface::class);
+        $this->manager->getInspector()->willReturn($this->inspector->reveal());
+        $this->hydrateEvent->getManager()->willReturn($this->manager->reveal());
 
         $this->subscriber = new PathSubscriber(
-            $this->inspector->reveal()
+            $this->manager->reveal()
```
danrot: This subscriber also doesn't take any constructor arguments anymore.

tests/Unit/Subscriber/Behavior/Path/AliasFilingSubscriberTest.php
```diff
@@ -83,11 +85,14 @@ public function setUp()
         $this->metadataFactory = $this->prophesize(MetadataFactoryInterface::class);
         $this->metadata = $this->prophesize(Metadata::class);
         $this->parentNode = $this->prophesize(NodeInterface::class);
+        $this->manager = $this->prophesize(DocumentManagerInterface::class);
```
danrot: Not quite sure, but I think it is not needed.

tests/Unit/Subscriber/Behavior/Path/AutoNameSubscriberTest.php
```diff
@@ -121,11 +122,17 @@ public function setUp()
         $this->nodeManager = $this->prophesize(NodeManager::class);
         $this->strategy = $this->prophesize(DocumentStrategyInterface::class);
 
+        $this->manager = $this->prophesize(DocumentManagerInterface::class);
```
danrot: Probably not needed.

tests/Unit/Subscriber/Behavior/Path/BasePathSubscriberTest.php
```diff
@@ -81,12 +70,13 @@ public function setUp()
         $this->parentDocument = new \stdClass();
         $this->nodeManager = $this->prophesize(NodeManager::class);
         $this->documentManager = $this->prophesize(DocumentManager::class);
-        $this->metadataFactory = $this->prophesize(MetadataFactoryInterface::class);
-        $this->metadata = $this->prophesize(Metadata::class);
         $this->parentNode = $this->prophesize(NodeInterface::class);
 
+        $this->manager = $this->prophesize(DocumentManagerInterface::class);
```
danrot: Probably not needed.

tests/Unit/Subscriber/Behavior/Path/ExplicitSubscriberTest.php
```diff
@@ -30,17 +31,19 @@ public function setUp()
         $this->configureEvent = $this->prophesize(ConfigureOptionsEvent::class);
         $this->parentNode = $this->prophesize(NodeInterface::class);
         $this->node = $this->prophesize(NodeInterface::class);
+        $this->manager = $this->prophesize(DocumentManagerInterface::class);
```
danrot: Probably not needed, and also missing all member variable declarations.

tests/Unit/Subscriber/Core/MappingSubscriberTest.php
```diff
@@ -43,11 +44,16 @@ public function setUp()
         $this->persistEvent->getLocale()->willReturn('de');
         $this->persistEvent->getAccessor()->willReturn($this->accessor);
 
+        $this->manager = $this->prophesize(DocumentManagerInterface::class);
```
danrot: Not needed, if the unused call of the `getManager` method of the `MappingSubscriber` is removed.

tests/Unit/Subscriber/Core/RegistratorSubscriberTest.php
```diff
@@ -59,15 +60,22 @@ class RegistratorSubscriberTest extends \PHPUnit_Framework_TestCase
     public function setUp()
     {
         $this->registry = $this->prophesize(DocumentRegistry::class);
-        $this->subscriber = new RegistratorSubscriber(
-            $this->registry->reveal()
-        );
+        $this->manager = $this->prophesize(DocumentManagerInterface::class);
```
danrot: Probably not needed.

tests/Unit/Subscriber/Phpcr/GeneralSubscriberTest.php
```diff
@@ -45,11 +47,26 @@ public function setUp()
         $this->document = new \stdClass();
         $this->node = $this->prophesize(NodeInterface::class);
 
-        $this->subscriber = new GeneralSubscriber(
-            $this->documentRegistry->reveal(),
-            $this->nodeManager->reveal(),
-            $this->eventDispatcher->reveal()
-        );
+        $this->manager = $this->prophesize(DocumentManagerInterface::class);
```
danrot: Probably not needed.

tests/Unit/Subscriber/Phpcr/RemoveSubscriberTest.php
```diff
@@ -32,10 +33,13 @@ public function setUp()
         $this->property1 = $this->prophesize(PropertyInterface::class);
         $this->property2 = $this->prophesize(PropertyInterface::class);
 
-        $this->subscriber = new RemoveSubscriber(
-            $this->documentRegistry->reveal(),
-            $this->nodeManager->reveal()
-        );
+        $this->manager = $this->prophesize(DocumentManagerInterface::class);
```
danrot: Probably not needed.

tests/Unit/Subscriber/Phpcr/ReorderSubscriberTest.php
```diff
@@ -34,10 +35,13 @@ public function setUp()
         $this->registry = $this->prophesize(DocumentRegistry::class);
         $this->event = $this->prophesize(ReorderEvent::class);
 
-        $this->subscriber = new ReorderSubscriber(
-            $this->nodeManager->reveal(),
-            $this->registry->reveal()
-        );
+        $this->manager = $this->prophesize(DocumentManagerInterface::class);
```
danrot, line 38: Also not needed, if unused method call is removed from `ReorderSubscriber`.

tests/Unit/Subscriber/Behavior/SubscriberTestCase.php
danrot: I would say YAGNI applies here. As long as we don't put any code in here I would not add another layer of inheritance. Also, we have a class like this in sulu/sulu, and I don't really like it... It's instantiating a few classes, which are not needed by all sub classes. I would prefer to also get rid of it there.

Summed up, I would not add empty classes to the inheritance hierarchy, because we might end up using it one time. If we do that all the time we would have a very bloated inheritance hierarchy.

lib/DocumentInspectorFactory.php
danrot, line 17: It is definitely not needed until now, and I don't want to introduce more changes than required. Makes stuff more complicated than it has to be. And if we merge it like that it is just there, without any additional value.

lib/DocumentManager.php
danrot, line 32: So this proxy document manager would implement the `_call` method, or how would you implement that?

lib/DocumentManagerInterface.php
danrot, line 130: No, this is the `DocumentManagerInterface`.

lib/EventDispatcher/DebugEventDispatcher.php
danrot, line 28: What if the `doDispatch` level exits early, e.g. because of an exception? I don't see where this gets reset afterwards.

lib/DocumentManager.php
danrot, line 276: Why? It is a private method, so it's only called by this class. And changing it the way I've suggested would have the same result, without catching `\Exception`, which only rarely makes sense, and having "flatter" code in this method.

lib/DocumentManager.php
```diff
@@ -175,4 +265,34 @@ private function getOptionsResolver($eventName)
...
+        if ($exception instanceof DocumentManagerException) {
+            // if the exception already has a document manager name, then it
+            // was thrown by this method in another document manager instance
+            // so we wrap it again.
+            //
+            // the user-facing exception class will be lost and we are assuming that standard
+            // exceptions (`DocumentNotFoundException`) will not be affected by this
+            // within the scope of normal usage.
+            if ($exception->getDocumentManagerName()) {
+                throw new DocumentManagerException($message, $this->context->getName(), $exception);
```
danrot, line 289: Why does there have to be a new exception created, if the given one already has a DocumentManagerName?

lib/DocumentManager.php
danrot, line 276: And throw phpdoc is missing.

lib/NodeManager.php
danrot: It's exactly the same message a 4 lines before...

lib/DocumentManagerContext.php
danrot, line 101: I see, but why don't we instantiate the `EventDispatcher` also in here then? Because it is the only not shared dependency being injected.

lib/Subscriber/Behavior/Path/ExplicitSubscriber.php
danrot, line 163: Can you check and remove if not necessary?

lib/DocumentManager.php
dantleech, line 32: Yes the `_call` method, which is the reason I did not implement it (also
the idea of having a `ExceptionWrappingDocumentManager` implementation).

lib/DocumentManager.php
dantleech, line 276: It means having to change things in multiple places in the future.

lib/Subscriber/Core/MappingSubscriber.php
```diff
@@ -274,11 +268,19 @@ private function hydrateReferenceField(
...
 
-            if ($referencedNode) {
-                $accessor->set(
+            try {
+                if ($referencedNode) {
+                    $accessor->set(
                     $fieldName,
-                    $this->proxyFactory->createProxyForNode($document, $referencedNode, $options)
+                    $proxyFactory->createProxyForNode($document, $referencedNode, $options)
                 );
```
danrot: Wrong indentation.

lib/DocumentManager.php
dantleech, line 289: So that you can see the document manager trace.

e.g.
1. You save a document in default DM.
2. The synchronization subscriber saves to the live DM.
3. Live DM throws an Exception.

If you didn't wrap the exception here then you would just see the
`default` manager name. But it didn't come from there.

With this change you will at least 2 exceptions, the first from the
default, the second from the live.

lib/Subscriber/Core/MappingSubscriber.php
danrot, line 278: But why not catching the `DocumentManagerException` again? The only throw I can encounter in the above code is a `RuntimeException` inheriting from the `DocumentManagerException`? We already had the same discussion 😕

lib/DocumentManager.php
danrot, line 276: What kind of things? The `DocumentManager` itself should only handle exceptions of type `DocumentManagerException` IMO anyway, so there is no need to change the method signature later.

lib/DocumentManagerContext.php
dantleech, line 101: - Event dispatcher is a service which needs subscribers to be added to it.
- Event dispatcher could be different (DebugEventDispatcher f.e.).
- Event dispatcher has no constructor dependencies.

We could make it optional. But I don't see any advantage at this stage.

lib/Subscriber/Core/MappingSubscriber.php
dantleech, line 278: It comes from the ProxyManager library.

lib/DocumentManager.php
danrot, line 289: Are you sure that's right? When the live DM is called last, it will add its own name to the exception. From what I can spot here it's exactly the opposite from what you are describing.
1. You save a document in default DM.
2. The synchronizatino subscriber saves to the live DM.
3. Live DM throws an exception
4. Live DM catches the exception and sets the name
5. Default DM also catches the Exception and sets the name again.

So, as it is currently implemented, you are ending up with the default name, although it has been failing in the live document manager.

lib/Subscriber/Core/MappingSubscriber.php
```diff
@@ -274,11 +268,19 @@ private function hydrateReferenceField(
...
+                }
+            } catch (\Exception $e) {
+                throw new RuntimeException(sprintf(
```
danrot, line 279: Import seems missing.

lib/Subscriber/Core/MappingSubscriber.php
danrot, line 278: What comes from the ProxyManager library? Certain exceptions? I am quite sure there is also a more specific exception for that.

lib/DocumentManager.php
dantleech, line 276: I don't really care. The point is that we repeat the same logic ~10
times in this class (catching the Exception) so anything we can do to
minimize the responsibility of each of those try/catch blocks the
better.

If you really feel strongly about this than it can be changed.

lib/Subscriber/Core/MappingSubscriber.php
dantleech, line 278: Again it doesn't matter what Exceptions. If there was an exception it
means that it could not hydrate the proxy class, which should not
happen. When it _does_ happen it is crucial to know on which field it
happened. This can save a lot of time when debugging.

If anybody is _listening_ for proxy manager exceptions, then somebody
has already done something very wrong.

lib/DocumentManager.php
danrot, line 289: Ok, missed that you are wrapping the exception here. It's ok for me like that.

lib/DocumentManager.php
danrot, line 276: I feel so strong about it. I also see it the other way round, this way the `DocumentManager` must also handle all other exceptions in some way (even if it only is about rethrowing them). The other way it only has to know about the `DocumentManagerException`.

lib/DocumentManagerContext.php
```diff
@@ -0,0 +1,199 @@
...
+     */
+    public function getSession()
+    {
+        return $this->session;
+    }
+
+    /**
+     * Return the event dispatcher instance.
+     *
+     * @return EventDispatcher
```
danrot: Import seems missing.

lib/DocumentManager.php
dantleech, line 276: An example I just discovered - when we encounter PHPCR exceptions we
probably want to catch them (Sulu should not know about the PHPCR
layer). As this method is here we can easily catch these exceptions in 
addition to the DocumentManager exceptions.

lib/DocumentManagerInterface.php
dantleech, line 130: We might be able to do that generally - providing that nothing outside of hte document manager needs it (not currently the case as it is required by the `ContentMapper`.

lib/ClassNameInflector.php
dantleech: Doesn't seem to be used.

lib/DocumentInspectorFactory.php
dantleech, line 17: > We did, but at this stage it requires more effort than it is worth, and
> then I am not sure if it is the best thing to do. Would rather wait
> before removing this, if at all.

lib/DocumentManager.php
dantleech, line 81: No.

lib/EventDispatcher/DebugEventDispatcher.php
dantleech, line 28: If there is an exception then the program is probably going to exit. This is just a debugging tool ...

lib/Subscriber/Behavior/Path/ExplicitSubscriber.php
dantleech, line 163: It looks like I introduced it early on - guessing it is todo with being able to persist the same document with different paths -- (the commit message was `implicit move`) -- this is required for the synchronization.

dantleech, line 163: It should be here because this class handles the `path` option, which is what this applies to.

lib/Subscriber/Core/MappingSubscriber.php
dantleech, line 279: Then the test seems to be missing.

tests/Functional/BaseTestCase.php
dantleech, line 49: Its the document manager. I think `manager` can be assumed to refer to the document manager in the document manager package.

danrot:
Replaced by #81

PR #69, opened 2016-03-11
Added options for remove events

wachterjohannes:
| Q | A |
| --- | --- |
| Bug fix? | no |
| New feature? | yes |
| BC breaks? | no |
| Deprecations? | no |
| Fixed tickets | none |
| Related issues/PRs | https://github.com/sulu-io/sulu/pull/2115 |
| License | MIT |
| Documentation PR | none |
#### What's in this PR?

This PR adds the possibility to pass options into the remove events. they could be used for example to determine which part of the document should be removed.
#### To Do
- [ ] Check if the documentation if valid
- [ ] CHANGELOG

### DIFF
--- a/lib/DocumentManager.php
+++ b/lib/DocumentManager.php
@@ -72,9 +72,11 @@ class DocumentManager implements DocumentManagerInterface
     /**
      * {@inheritdoc}
      */
-    public function remove($document)
+    public function remove($document, array $options = [])
     {
-        $event = new Event\RemoveEvent($document);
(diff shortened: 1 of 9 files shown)


dantleech:
As previously stated. I do not believe that changing the meaning of `remove` here is a valid thing to do. I can only suggest adding a `->removeLocale` method to the API currently.

wachterjohannes:
@dantleech this PR is independent from the discussion we have today

dantleech:
Then what options do you propose to add?

PR #72, opened 2016-03-15
Fixed exception swallowing for invalid locales.

### LINKED ISSUE #70: InvalidLocaleException is caught and discarded in MappingSubscriber
The mapping subscriber has a [series of try catches](https://github.com/sulu-io/sulu-document-manager/blob/master/lib/Subscriber/Core/MappingSubscriber.php#L132) which swallow an `InvalidLocaleBehavior`.

Consuming exceptions is a bad practice, and this WILL cause the system to behave an an unexpected manner when such an exception is produced, as the mapping will not be set as intended.

dantleech:
Fixes #70

This PR replaces the hack used to ignore invalid locales when given to the property encoder.

Instead this PR changes the logic to use the default locale when no locale is set when persisting.


lib/Exception/InvalidLocaleException.php
```diff
@@ -1,37 +0,0 @@
...
- * This source file is subject to the MIT license that is bundled
- * with this source code in the file LICENSE.
- */
-
-namespace Sulu\Component\DocumentManager\Exception;
-
-/**
- * Indicates invalid locale argument.
- */
-class InvalidLocaleException extends \InvalidArgumentException
```
dantleech, line 16: Exception not needed IMO as we should never have to catch it.

lib/Exception/InvalidLocaleException.php
```diff
@@ -1,37 +0,0 @@
...
-{
-    /**
-     * @var string
-     */
-    private $locale;
-
-    public function __construct($locale)
```
dantleech, line 23: This exception was only called when $locale was null..

lib/Subscriber/Core/RegistratorSubscriber.php
```diff
@@ -83,7 +84,7 @@ public function configureOptions(ConfigureOptionsEvent $event)
      *
      * @param HydrateEvent $event
      */
-    public function handleDefaultLocale(HydrateEvent $event)
+    public function handleDefaultLocale(AbstractMappingEvent $event)
```
dantleech, line 87: The `hydrate` event already handled the default locale, it was just missing the persist event.

tests/Unit/PropertyEncoderTest.php
```diff
@@ -60,42 +60,24 @@ public function testEncodeSystem()
...
-
-    /**
-     * It should throw exception.
-     */
-    public function testLocalizedContentEmptyLocale()
+    public function provideEncodeLocalizedException()
     {
-        $this->setExpectedException(\InvalidArgumentException::class);
-
-        $this->encoder->encode('content_localized', 'test', null);
```
dantleech, line 89: This test was the same as the previous one.

tests/Unit/PropertyEncoderTest.php
```diff
@@ -60,42 +60,24 @@ public function testEncodeSystem()
...
-    }
-
-    /**
-     * It should throw exception.
```
dantleech, line 93: When should it have thrown an exception?

danrot:
changelog is missing

danrot:
Tests are failing... and to what issue in Sulu does this relate?

dantleech:
It doesn't relate to any specific issue, it is just not very good code - we should not swallow exceptions and leave the system in an inconsistent state.

It does however relate to the concerns you had about persisting without a locale - it seems to me this try/catch stuff was an attempt to deal with that.

lib/Subscriber/Core/RegistratorSubscriber.php
danrot, line 87: I still don't like the concept of a default locale when persisting, and Sulu already relies on not passing a locale in certain situations. I guess this change would make all the subscribers checking on the locale changing the behavior? Is this change necessary to these try/catches in the `MappingSubscriber`?

dantleech, line 87: To be honest I really don't know anymore what the policy is with default
locales. The system was initially required a locale. I thought I then
implemented a default locale (which is the same behavior as the
PHPCR-ODM and the Doctrine extensions IIRC).
- The document should be persisted in the same locale that it was loaded
  in, if no locale is given when persisting.
- If it is a new document, then the default locale should be used.
- If the document is not localized, then having a default locale does
  not matter.

danrot, line 87: I can only tell you that this is probably a change breaking Sulu in its current state. As already said Sulu relies on this behavior, since we are saving stuff from time to time which is not locale related (like permissions). Therefore it is critical for me that this behavior does not change.

dantleech, line 87: The point is that you map a field to use localized property name encoding. If you save without a locale, here, then any multi language property you have set will not be persisted.

Swallowing exceptions is not the way to handle this, the logic, whatever it is, should b explicit, you can see that this is bad as the exception swallowing behavior is duplicated multiple times.

For the sake of simplicty and clarity and predictability I say that the locale should be mandatory, defauting to the default locale. This saves lots of trouble and will make the system more stable - this is the way the PHPCR-ODM works, and I would bet it is how the Doctrine Extensions work also.

If you load a document, it will be loaded in the default locale. If you save it (without a locale) it would be saved in that same locale.

What issues arrise by using the default locale?

I appreciate that none of you _like_ this behavior, but that doesn't make it invalid.

danrot, line 87: The fact that we use [checks like this](https://github.com/sulu/sulu/blob/7f745ac7c5617bfcf3a92de53871e03829c08a19/src/Sulu/Component/Content/Document/Subscriber/StructureSubscriber.php#L202) makes this change of the document manager incompatible with the current state of Sulu. Also, it still doesn't make any sense to pass a locale when we e.g. move a page, since this change is valid for all localizations, and not only a single one. This is just confusing and screams for unexpected behavior.

dantleech, line 87: Not having a locale at all means every thing needing to consider the
abscence of the locale. If you take for granted the presence of the
locale, then you have no problem.

What is the issue? If you use a default locale you do not need to make
checks like "that". Everybody is happy.

danrot, line 87: Not the people concerned about performance, since an entire language es saved, when it is not required at all.

dantleech, line 87: This is where you need a change manager to act upon the
MappingSubscriber (which SHOULD be where all the mapping happens) so as
to only write what has not changed.

However that would have only a marginal effect, the big issue would come
from the legact content system which needs to be decoupled from the
document manager as previously discussed.

PR #74, opened 2016-04-15
Document context context

dantleech:
(re) introduced document manager context.


lib/DocumentManager.php
```diff
@@ -76,25 +59,25 @@ public function __construct(
...
         // instantiate other objects scoped to this document manager.
-        $this->nodeManager = new NodeManager($session);
-        $this->proxyFactory = new ProxyFactory($this, $lazyProxyFactory, $metadataFactory);
-        $this->registry = new DocumentRegistry($defaultLocale);
+        $nodeManager = new NodeManager($session);
+        $proxyFactory = new ProxyFactory($this, $lazyProxyFactory, $metadataFactory);
+        $documentRegistry = new DocumentRegistry($defaultLocale);
+
+        $this->context = new DocumentManagerContext(
+            $this,
```
dantleech: DocumentManagerContext contains an instance of the document manager which instantiatates it.

lib/ProxyFactory.php
```diff
@@ -78,9 +78,9 @@ public function createProxyForNode($fromDocument, NodeInterface $targetNode, $op
 
             // If the parent is not loaded in the correct locale, reload it in the correct locale.
             if ($registry->getOriginalLocaleForDocument($document) !== $locale) {
-                $hydrateEvent = new HydrateEvent($this->documentManager, $targetNode, $locale);
+                $hydrateEvent = new HydrateEvent($this->documentManager->getContext(), $targetNode, $locale);
                 $hydrateEvent->setDocument($document);
-                $this->documentManager->getEventDispatcher()->dispatch(Events::HYDRATE, $hydrateEvent);
+                $this->documentManager->getContext()->getEventDispatcher()->dispatch(Events::HYDRATE, $hydrateEvent);
```
dantleech: Could probably remove the document manager dep. here.

dantleech:
I removed this (document manager context) initially because it meant having an artificial wrapper around the dependencies of the document manager - a wrapper which accepts an instance of the document manager (containing the its dependencies) and then the dependencies of the document manager.

The event has been modfied to provide direct accesors to the document manager services, which are proxies to the DocumentManagerContext object passed to it, which was instantiated by the document manager.

dantleech:
An alternative approach would be to:
- Instantiate the `DocumentContext` before the `DocumentManager`
- The `DocumentContext` would produce the `DocumetManager` instance.
- The `DocumentManager` would then only have only "external "API methods and not reveal its insides.
- The `DocumentManagerRegistry` would then have both `getManager` and `getContext` methods, so that the user could at least be able to access the "insides" of the DocumetnManager when the need arises.

But I am again rather more on the side of keeping it as it already is in `document_context` branch.

/cc @chirimoya @danrot

lib/DocumentInspectorFactoryInterface.php
```diff
@@ -11,27 +11,29 @@
...
-     * manager.
+     * context.
      *
-     * @param DocumentManagerInterface
+     * @param DocumentManagerContext
      *
      * @return DocumentInspector
      */
-    public function getInspector(DocumentManagerInterface $manager);
+    public function getInspector(DocumentManagerContext $context);
```
danrot, line 36: Shouldn't the `$context` just offer a `getDocumentInspector` method? Having a factory here feels over engineered for me.

lib/DocumentManager.php
```diff
@@ -76,25 +59,25 @@ public function __construct(
...
-
         // instantiate other objects scoped to this document manager.
-        $this->nodeManager = new NodeManager($session);
-        $this->proxyFactory = new ProxyFactory($this, $lazyProxyFactory, $metadataFactory);
-        $this->registry = new DocumentRegistry($defaultLocale);
+        $nodeManager = new NodeManager($session);
+        $proxyFactory = new ProxyFactory($this, $lazyProxyFactory, $metadataFactory);
+        $documentRegistry = new DocumentRegistry($defaultLocale);
+
+        $this->context = new DocumentManagerContext(
```
danrot: Why don't we create all these classes in the `Extension` class? Then we could probably even remove the `getContext` method from this class.

lib/DocumentInspectorFactoryInterface.php
dantleech, line 36: The factory is required to enable the end user to use a custom inspector.

lib/DocumentManagerContext.php
```diff
@@ -0,0 +1,77 @@
...
+
+class DocumentManagerContext
+{
+    private $documentManager;
+    private $nodeManager;
+    private $documentRegistry;
+    private $proxyFactory;
+    private $eventDispatcher;
+    private $metadataFactory;
+    private $session;
```
danrot, line 57: PHPDocs missing

lib/DocumentManagerContext.php
```diff
@@ -0,0 +1,77 @@
...
+    public function __construct(
+        DocumentManagerInterface $documentManager,
+        NodeManager $nodeManager,
+        DocumentRegistry $documentRegistry,
+        ProxyFactory $proxyFactory,
+        EventDispatcher $eventDispatcher,
+        MetadataFactoryInterface $metadataFactory,
+        SessionInterface $session
+    )
+    {
```
danrot: previous two lines into one

lib/DocumentInspectorFactoryInterface.php
danrot, line 36: We also did that until now without a factory, didn't we?

dantleech, line 36: Yes, because we instantiated a single instance.

lib/Event/AbstractDocumentManagerContextEvent.php
```diff
@@ -0,0 +1,88 @@
...
+        return $this->getContext()->getSession();
+    }
+
+    /**
+     * @return string
+     */
+    public function getDebugMessage()
+    {
+        return '';
+    }
```
danrot, line 85: This is exactly the same as the only method in the `AbstractEvent`, so I would remove it. This way the chain of inheritance already includes 5 classes.

lib/DocumentInspectorFactoryInterface.php
danrot, line 36: Yeah, and we still don't need more than a single instance right? I think YAGNI should apply here.

dantleech, line 36: No - each document manager needs its own instance of the inspector.

danrot:
I like the alternative approach you've suggested here even better. In an ideal world I'd say that both, the `DocumentManager` and the `DocumentManagerRegistry` don't have a `getDocumentContext` method. The context should only be accessed by the `EventSubscriber`, if possible. If necessary I would only put it on the `DocumentManagerRegistry`, since the average user will be using the `DocumentManager` and does not have to struggle with these concepts anyway, making future refactorings at least a little bit easier.

dantleech:
ok, will in that case the Context would basically become a hard-coded
container for the dependencies. 

We would either need to pass both the Context and the DocumentManager to
the events: `new FooEvent($manager, $context)` which is a potential
source of error, or we would make the `Context` instantiate (its)
DocumentManager (and create the DM through a factory service in the DI
config).

I would also remove the getters from the Event (getProxyManager,
getMetadtaFactory, etc) retaining "getContext" and perhaps "getManager".

It is again quite alot of work to implement (at least 1 day - this took
half a day and then with the Sulu side it is going to to longer again).

I would still err on the side of simplicity and keep public methods in
the DocumentManager and mark them as `@internal`, they would not be part
of the `DocumentManagerInterface`.

danrot:
So what you are saying is that we would have to pass the manager and context to the events, if the context is not created in combination with the manager by a factory?

Why removing the getters again? I think that's a nice solution, follows the law of demeter, and makes it easier for the developer programming the subscriber.

Putting the methods only on the class and not on the interface would mean that we can't use the interface in type hints, what I don't really like...

dantleech:
On Mon, Apr 18, 2016 at 12:44:05AM -0700, Daniel Rotter wrote:
>    So what you are saying is that we would have to pass the manager and
>    context to the events, if the context is not created in combination with
>    the manager by a factory?

The subscribers need access to both the manager and the context
(technically we should be able to avoid subscribers accessing the manager
however, they have the EventDispatcher via. the context after all).

I would say to reduce contamination the Context should be the factory
that creates the DocumentManager.

> 
>    Why removing the getters again? I think that's a nice solution, follows
>    the law of demeter, and makes it easier for the developer programming the
>    subscriber.

It also violates SPOT. There are pros and cons.

> 
>    Putting the methods only on the class and not on the interface would mean
>    that we can't use the interface in type hints, what I don't really like...

We would create another interface: `DocuemntManagerServiceInterface` for
example.

danrot:
I don't see how we should need a new interface here. If we move the instantiation to the `Extension` class and have a `getManager` and `getContext` method to the `DocumentManagerRegistry` we are all set, aren't we?

lib/DocumentInspectorFactoryInterface.php
danrot, line 36: Yeah - and if we put that instance in the context, we don't need a factory for this.

dantleech, line 36: It means we need to use an abstract definition for the Inspector (which
is ok) and that the DI container has quite some responsiblity in
correctly instantiating all the required services with the correct
scope, instead of that responsibility being contained (and tested)
within the document manager itself.

danrot, line 36: The problem for me is that it adds complexity and inconsistency. The same argument could apply for other parts of the context as well, and we don't introduce factories for all of them. Also, some of the classes are created in the `DocumentManager`'s constructor. For me that's very confusing, and takes some time to understand, and people won't remember it. If we put all of it in the `Extension`, there is a single place we know we have to look this kind of stuff up. And this functionality can also be tested in this bundle.

dantleech:
@danrot the `DocumentManagerContext` is now injected into the `DocumentManager`, and distributed with the events.

need to refactor the `multiple_document_managers` branch in Sulu to meet this change.

dantleech:
Remove DocumentInspectorFactory

lib/DocumentInspectorFactoryInterface.php
dantleech, line 36: After looking at the DI config, for this moment I am going to keep the factory. It makes the most sense given the _current_ configuration, instantiating the `DocumentInspector` in the DI is just using a different type of factory, and it means using the DocumentContext as a factory service to get various dependencies.

We could maybe change this later if we change the way the document manager services `DocumentRegistry` and `NodeManager` are instantiated.

PR #75, opened 2016-04-18
Added name option

dantleech:
This PR allows document managers to be identified with a name and any exceptions thrown through the primary API methods will be either wrapped with a `DocumentManagerException`.

If the Exception is already an instance of `DocumentManagerException` (f.e. a `DocumentNotFoundException`) then the document name will be set on that exception itself.

### DIFF
--- a/lib/DocumentManager.php
+++ b/lib/DocumentManager.php
@@ -13,6 +13,7 @@ namespace Sulu\Component\DocumentManager;
 
 use PHPCR\SessionInterface;
 use ProxyManager\Factory\LazyLoadingGhostFactory;
+use Sulu\Component\DocumentManager\Exception\DocumentManagerException;
 use Symfony\Component\EventDispatcher\EventDispatcherInterface;
 use Symfony\Component\OptionsResolver\OptionsResolver;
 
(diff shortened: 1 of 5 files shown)


dantleech:
Note that this is a PR on the existing `document_context` branch and I will merge it when the tests are green.

lib/DocumentManager.php
```diff
@@ -69,6 +69,7 @@ class DocumentManager implements DocumentManagerInterface
     private $session;
 
     public function __construct(
```
dantleech, line 83: Add docs

lib/DocumentRegistry.php
```diff
@@ -60,10 +60,16 @@ class DocumentRegistry
...
     /**
+     * @var string
+     */
+    private $name;
+
+    /**
      * @param $defaultLocale
      */
-    public function __construct($defaultLocale)
+    public function __construct($name, $defaultLocale)
```
dantleech: Doc

lib/DocumentManager.php
```diff
@@ -25,6 +26,12 @@
  *
  * This class is therefore an event dispatcher and a service container and has
  * no other business logic within it.
+ *
+ * All primary methods are wrapped in a try catch block, the exception is then
+ * wrapped in a `DocumentManagerException` with a reference to this document
+ * managers name.
+ *
+ * TODO: Create a proxy document manager which automatically wraps all calls.
```
dantleech, line 34: Rather than wrap the methods we could create a proxy document manager (`ExceptionCatchingDocumentManager`) which wraps _all_ calls with th exception handling logic.

note that this is better in that we reduce duplication, increase test coverage (we only need to test one case) but increase magic by using `__call`

lib/DocumentManager.php
```diff
@@ -97,126 +109,179 @@ public function __construct(
         $this->registry = new DocumentRegistry($defaultLocale);
     }
 
+    public function setName($name)
```
dantleech, line 118: Doc.

PR #81, opened 2016-05-04
added publish method to document manager

danrot:
| Q | A |
| --- | --- |
| Bug fix? | no |
| New feature? | yes |
| BC breaks? | no |
| Deprecations? | no |
| Fixed tickets | fixes #2028 |
| Related issues/PRs | #issuenum |
| License | MIT |
| Documentation PR | TODO |
#### What's in this PR?

This PR adds a publish method to the `DocumentManager`.
#### Why?

Because we have to publish documents for the drafting feature.


lib/DocumentManagerInterface.php
```diff
@@ -60,7 +60,7 @@ public function remove($document);
      * Move the PHPCR node to which the document is mapped to be a child of the node at the given path or UUID.
      *
      * @param object $document
-     * @param string $destId Path or UUID
```
danrot, line 63: Maybe we should change that to only path... The problem is that not every node has a UUID (e.g. or maybe only the newly created ones), but every node has a path (also the ones which are only existing in the current session).

chirimoya, line 63: ???

lib/Event/PublishEvent.php
```diff
@@ -0,0 +1,39 @@
...
+    public function __construct($document, $locale)
+    {
+        $this->document = $document;
+        $this->locale = $locale;
+    }
+
+    /**
+     * Sets the node this event should operate on.
+     *
+     * TODO Check if should be move to DocumentManager itself
```
chirimoya: issue

chirimoya:
ACK

lib/Event/PublishEvent.php
danrot: https://github.com/sulu/sulu-document-manager/issues/94

lib/DocumentManagerInterface.php
danrot, line 63: Fixed comment

PR #82, opened 2016-05-06
Introduced mandatory locale in document-registry

wachterjohannes:
This PR introduces the mandatory locale in the registry. This will fix the problem with "rehydrating" and same document for different locales.

### DIFF
--- a/lib/Behavior/Mapping/LocaleBehavior.php
+++ b/lib/Behavior/Mapping/LocaleBehavior.php
@@ -29,4 +29,18 @@ interface LocaleBehavior
      * @param string $locale
      */
     public function setLocale($locale);
+
+    /**
+     * Return the documents original locale.
+     *
+     * @return string
+     */
+    public function getOriginalLocale();
+
+    /**
+     * Sets the document original locale.
+     *
+     * @param string $locale
+     */
+    public function setOriginalLocale($locale);
 }
--- a/lib/DocumentRegistry.php
+++ b/lib/DocumentRegistry.php
@@ -84,7 +84,7 @@ class DocumentRegistry
         $this->documentMap[$oid] = $document;
         $this->documentNodeMap[$oid] = $uuid;
         $this->nodeMap[$node->getIdentifier()] = $node;
-        $this->nodeDocumentMap[sprintf('%s-%s', $node->getIdentifier(), $locale)] = $document;
+        $this->nodeDocumentMap[$this->getNodeLocaleKey($node->getIdentifier(), $locale)] = $document;
         $this->documentLocaleMap[$oid] = $locale;
     }
 
@@ -112,7 +112,7 @@ class DocumentRegistry
      */
     public function hasNode(NodeInterface $node, $locale)
     {
-        return array_key_exists(sprintf('%s-%s', $node->getIdentifier(), $locale), $this->nodeDocumentMap);
+        return array_key_exists($this->getNodeLocaleKey($node->getIdentifier(), $locale), $this->nodeDocumentMap);
     }
 
     /**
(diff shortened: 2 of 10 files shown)


lib/DocumentRegistry.php
```diff
@@ -73,44 +69,22 @@ public function __construct($defaultLocale)
...
-     *
-     * @param object $document
-     * @param string $locale
-     * @param null|string $originalLocale
-     */
-    public function updateLocale($document, $locale, $originalLocale = null)
-    {
-        $oid = $this->getObjectIdentifier($document);
-        $this->originalLocaleMap[$oid] = $originalLocale;
+        $this->nodeDocumentMap[sprintf('%s-%s', $node->getIdentifier(), $locale)] = $document;
```
danrot: Any chances of naming collisions here? Maybe use a different sign than `-` to avoid confusion with the dashes from the UUID? `_` comes to mind.

lib/DocumentRegistry.php
```diff
@@ -132,12 +106,13 @@ public function hasDocument($document)
...
      * @param NodeInterface $node
+     * @param string $locale
      *
      * @return bool
      */
-    public function hasNode(NodeInterface $node)
+    public function hasNode(NodeInterface $node, $locale)
     {
-        return isset($this->nodeDocumentMap[$node->getIdentifier()]);
+        return array_key_exists(sprintf('%s-%s', $node->getIdentifier(), $locale), $this->nodeDocumentMap);
```
danrot: The `sprintf` call should probably be an own method, is already used above.

lib/DocumentRegistry.php
```diff
@@ -232,15 +203,16 @@ public function getOriginalLocaleForDocument($document)
...
+     * @return mixed If the node is not managed
      */
-    public function getDocumentForNode(NodeInterface $node)
+    public function getDocumentForNode(NodeInterface $node, $locale)
     {
         $identifier = $node->getIdentifier();
         $this->assertNodeExists($identifier);
 
-        return $this->nodeDocumentMap[$identifier];
+        return $this->nodeDocumentMap[sprintf('%s-%s', $identifier, $locale)];
```
danrot: Same as above, extract an own method.

lib/Subscriber/Core/RegistratorSubscriber.php
```diff
@@ -144,11 +143,7 @@ public function handleStopPropagationAndResetLocale(HydrateEvent $event)
             (true === $this->documentRegistry->isHydrated($document) && $originalLocale === $locale)
         ) {
             $event->stopPropagation();
-
-            return;
         }
-
-        $this->documentRegistry->updateLocale($document, $locale, $locale);
```
danrot, line 151: Shouldn't the php doc of this method be updated because of this?

lib/Behavior/Mapping/LocaleBehavior.php
```diff
@@ -29,4 +29,18 @@ public function getLocale();
...
+     * @return string
+     */
+    public function getOriginalLocale();
+
+    /**
+     * Sets the document original locale.
+     *
+     * @param string $locale
+     */
+    public function setOriginalLocale($locale);
```
danrot, line 45: I am quite skeptical about this change... The original locale makes only sense for documents with the `ShadowLocaleBehavior`, doesn't it? This behavior already has a `shadowLocale`, which would be duplicating information then.

wachterjohannes, line 45: also for ghost pages this locale will be used. this will also be different to locale for snippets if there ghost content is loaded.

lib/Subscriber/Core/RegistratorSubscriber.php
wachterjohannes, line 151: fixed

wachterjohannes:
closed in favor of https://github.com/sulu/sulu-document-manager/pull/88

PR #83, opened 2016-05-11
Fixed auto-name subscriber to rename at the very end of persist

wachterjohannes:
This PR fixes the `AutoNameSubscriber` to rename the node at the very en of persist.

Newly created issue with phpbench https://github.com/sulu/sulu-document-manager/issues/84

### DIFF
--- a/lib/NameResolver.php
+++ b/lib/NameResolver.php
@@ -27,21 +27,11 @@ class NameResolver
      */
     public function resolveName(NodeInterface $parentNode, $name, $forNode = null)
     {
-        if ($forNode && $parentNode->hasNode($name) && $parentNode->getNode($name) == $forNode) {
-            return $name;
-        }
-
         $index = 0;
         $baseName = $name;
-        do {
-            if ($index > 0) {
-                $name = $baseName . '-' . $index;
-            }
-
-            $hasChild = $parentNode->hasNode($name);
-
-            ++$index;
-        } while ($hasChild);
+        while ($parentNode->hasNode($name) && (!$forNode || $parentNode->getNode($name) !== $forNode)) {
+            $name = $baseName . '-' . ++$index;
+        }
 
         return $name;
     }
--- a/lib/Subscriber/Behavior/Path/AutoNameSubscriber.php
+++ b/lib/Subscriber/Behavior/Path/AutoNameSubscriber.php
@@ -87,7 +87,12 @@ class AutoNameSubscriber implements EventSubscriberInterface
     {
         return [
             Events::CONFIGURE_OPTIONS => 'configureOptions',
-            Events::PERSIST => ['handlePersist', 480],
+            Events::PERSIST => [
+                ['handlePersist', 480],
+                // the rename has to done at the very end to avoid ItemNotFoundException
+                // see https://github.com/jackalope/jackalope/blob/3cf6e0582acb26b5b83b3445be238ea8aadf46ec/src/Jackalope/ObjectManager.php#L429
+                ['handleRename', -480],
+            ],
             Events::MOVE => ['handleMove', 480],
             Events::COPY => ['handleCopy', 480],
         ];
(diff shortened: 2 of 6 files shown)


lib/Subscriber/Behavior/Path/AutoNameSubscriber.php
```diff
@@ -125,16 +130,57 @@ public function handleCopy(CopyEvent $event)
...
+    /**
+     * Renames node if necessary.
+     *
+     * @param PersistEvent $event
+     */
+    public function handleRename(PersistEvent $event)
+    {
         $document = $event->getDocument();
+        $defaultLocale = $this->registry->getDefaultLocale();
+        if (!$event->getOption('auto_name')
```
danrot, line 154: New line before this line.

wachterjohannes:
@danrot fixed

tests/Unit/Subscriber/Behavior/Path/AutoNameSubscriberTest.php
```diff
@@ -187,7 +190,22 @@ public function testAlreadyHasNode()
...
+        $this->node->hasNode()->willReturn(true);
+        $this->node->getName()->willReturn('foo');
+
+        $this->subscriber->handlePersist($this->persistEvent->reveal());
+    }
+
+    /**
+     * It should rename the node if the document is being saved in the default locale.
+     */
+    public function test()
```
danrot: That's not a very good method name 😉

tests/Unit/Subscriber/Behavior/Path/AutoNameSubscriberTest.php
```diff
@@ -144,15 +145,16 @@ public function testNotInstanceOfAutoName()
...
      */
     public function testNoTitle()
     {
+        $this->setExpectedException(DocumentManagerException::class);
+
         $this->persistEvent->hasNode()->willReturn(false);
         $this->document->getTitle()->willReturn(null);
         $this->persistEvent->getOption('auto_name')->willReturn(true);
         $this->persistEvent->getDocument()->willReturn($this->document->reveal());
+        $this->persistEvent->getParentNode()->willReturn($this->parentNode);
```
danrot: `reveal` missing?

tests/Unit/Subscriber/Behavior/Path/AutoNameSubscriberTest.php
wachterjohannes: that seems to be a mistake because this test is the same as the test before.

tests/Unit/Subscriber/Behavior/Path/AutoNameSubscriberTest.php
wachterjohannes: fixed

tests/Unit/Subscriber/Behavior/Path/AutoNameSubscriberTest.php
```diff
@@ -171,6 +173,7 @@ public function testAutoName()
     public function testAutoNameWithDisabledOption()
     {
         $this->persistEvent->getOption('auto_name')->willReturn(false);
+        $this->persistEvent->getDocument()->willReturn(null);
```
danrot: Shouldn't a document of type `AutonameBehavior` be returned here? Otherwise the test would also succeed if  `auto_name` is not set to false.

wachterjohannes: Thats right (:

danrot:
I would also add a function test for that.

lib/NameResolver.php
```diff
@@ -27,21 +27,11 @@ class NameResolver
      */
     public function resolveName(NodeInterface $parentNode, $name, $forNode = null)
     {
-        if ($forNode && $parentNode->hasNode($name) && $parentNode->getNode($name) == $forNode) {
-            return $name;
-        }
-
-        $index = 0;
+        $index = 1;
```
danrot: I would leave that one with `0`, and use `++$index` in the loop.

danrot:
Is that now ready to review? Because you've said yesterday that you were still experiencing a bug.

wachterjohannes:
when you sulu tests are green this is also good!

lib/NameResolver.php
wachterjohannes: fixed

PR #86, opened 2016-05-12
Fixed resolving name for already incremented node names

danrot:
This PR fixes the resolving of node names. When the `NameResolver::resolveName` got passed a `$forNode` which was already suffixed with an increment (e.g. `foo-1`) it didn't recognize it as the same node. That means that saving a page with an already incremented node name always switched between two increments (`foo-1` gets `foo-2` when being saved, and turns back to `foo-1` when saved another time).

This issue made some problems in combination with the publishing, because it moves a node in the session. Since this is neither necessary nor the desired behavior, changing this seems like a valid fix.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@ CHANGELOG for Sulu
 ==================
 
 * dev-develop
+    * BUGFIX      #86 Fixed resolving name for already incremented node names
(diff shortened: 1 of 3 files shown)


wachterjohannes:
@danrot i have included this into the hotfix PR https://github.com/sulu/sulu-document-manager/pull/83/files#diff-8ddc58cf5731ab79d00888d92c84a8d7R32

PR #91, opened 2016-06-08
Added replacer to avoid 10e in node-names

wachterjohannes:
This PR avoids `10e` in node names by replacing it. This replaces https://github.com/sulu/sulu/pull/2387 .

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,9 @@
 CHANGELOG for Sulu
 ==================
 
+* dev-master
+    * HOTFIX      #91 Added replacer to avoid 10e in node-names
+
 * 0.8.0 (2016-07-28)
     * FEATURE     #94 Added removeDraft method to DocumentManager
 
--- a/lib/Subscriber/Behavior/Path/AutoNameSubscriber.php
+++ b/lib/Subscriber/Behavior/Path/AutoNameSubscriber.php
@@ -199,6 +199,9 @@ class AutoNameSubscriber implements EventSubscriberInterface
 
         $name = $this->slugifier->slugify($title);
 
+        // jackrabbit can not handle node-names which contains a number followed by "e" e.g. 10e
+        $name = preg_replace('((\d+)([eE]))', '$1-$2', $name);
+
         return $this->resolver->resolveName($parentNode, $name, $node, $autoRename);
     }
 
--- a/tests/Unit/Subscriber/Behavior/Path/AutoNameSubscriberTest.php
+++ b/tests/Unit/Subscriber/Behavior/Path/AutoNameSubscriberTest.php
@@ -160,6 +160,36 @@ class AutoNameSubscriberTest extends \PHPUnit_Framework_TestCase
         $this->subscriber->handlePersist($this->persistEvent->reveal());
     }
 
+    public function provide10eData()
+    {
+        return [
+            ['10e', '10-e'],
+            ['.10e', '.10-e'],
+            ['-10e', '-10-e'],
+            ['%10e', '%10-e'],
+            ['test-10e-name', 'test-10-e-name'],
+            ['test.10e-name', 'test.10-e-name'],
+            ['test%10e-name', 'test%10-e-name'],
+            ['test-10E-name', 'test-10-E-name'],
+            ['test.10E-name', 'test.10-E-name'],
+            ['test%10E-name', 'test%10-E-name'],
+            ['test10E-name', 'test10-E-name'],
+            ['test-9e-name', 'test-9-e-name'],
+            ['test-500e-name', 'test-500-e-name'],
+        ];
+    }
+
+    /**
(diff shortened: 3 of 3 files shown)


lib/Subscriber/Behavior/Path/AutoNameSubscriber.php
```diff
@@ -196,6 +196,9 @@ private function getName(AutoNameBehavior $document, NodeInterface $parentNode,
 
         $name = $this->slugifier->slugify($title);
 
+        // jackrabbit can not handle node-names which contains a number followed by "e" e.g. 10e
+        $name = preg_replace('((\d+)e)', '$1-e', $name);
```
dbu: i think in my analysis at https://github.com/jackalope/jackalope/issues/313 i found that only .-% before a number trigger the problem. and that uppercase E is also a problem. maybe `'([\.-%]\d+)([eE])', '$1-$2` ?

dbu:
can the user ever manually define node names? or does that also go through this auto-namer? you might want to un-escape when loading, unless the names are never publicly visible (not even in api calls). if you do, you should replace (\d)+(-+)([eE]) by $1$2-$3 and always remove one `-` so that names that would need no escaping are not mangled.

wachterjohannes:
@dbu all nodes which we use for querying is generated by this auto-name behavior. i don't think we want to undo this for the user because if he want to use this path it should be valid.

lib/Subscriber/Behavior/Path/AutoNameSubscriber.php
wachterjohannes: thats right (: i will change it to this regex. thanks for your feedback

dbu: np

danrot:
The JIRA issue is https://issues.apache.org/jira/browse/JCR-3985, just putting it here for reference.

danrot:
This error also happens for `.10e` and only `10e`, but the regex doesn't cover these. I guess when it comes to the regex they are both the same, because the `.` is probably removed before.

dbu:
the jira ticket says that its about big decimal parsing. which makes no
sense probably, but maybe reading how java parses big decimals can help
with the escaping to cover all cases.

wachterjohannes:
@danrot most of the cases should be covered because `.` or `,` will be replaced by a `-` so they wont ma a problem there. i have tested now multiple formats and i have not found any which causes this problem ...

danrot:
The problem is that the `.` doesn't come to the regex, at least that is my guess. And just calling a site `10e` also didn't work, which makes sense, since the regex says that there has to be something before `10e`.

wachterjohannes:
@danrot fixed

CHANGELOG.md
```diff
@@ -10,6 +11,10 @@ CHANGELOG for Sulu
     * BUGFIX      #88 Introduced mandatory locale in document-registry
     * FEATURE     #81 Added publish method to DocumentManager
     * BUGFIX      #85 Fixed get locale for proxy
+=======
+* dev-master
+    * HOTFIX      #91 Added replacer to avoid 10e in node-names
+>>>>>>> added replacer to avoid 10e in node-names
```
danrot: Please solve that merge conflict 😃

wachterjohannes: haha (: seems that i need glasses

wachterjohannes:
@danrot now everything should be fine

lib/Subscriber/Behavior/Path/AutoNameSubscriber.php
```diff
@@ -7,8 +7,10 @@
  *
  * This source file is subject to the MIT license that is bundled
  * with this source code in the file LICENSE.
+ *
  */
 
+
```
danrot: These two additions were not here before, were they? Both of them should be removed, same for the other file.

danrot:
And can you please rebase the branch and remove the merge commit? 😃

chirimoya:
ping @wachterjohannes

wachterjohannes:
@danrot rebased

lib/Subscriber/Behavior/Path/AutoNameSubscriber.php
```diff
@@ -199,6 +200,9 @@ private function getName(
 
         $name = $this->slugifier->slugify($title);
 
+        // jackrabbit can not handle node-names which contains a number followed by "e" e.g. 10e
+        $name = preg_replace('(([\.%-]?\d+)([eE]))', '$1-$2', $name);
```
danrot: The first part of the regex `[\.%-]?` could be removed now, since it matches everytime. The `?` modifier says that it can occur, but doesn't have to. If we really want to make this happen only when there is a `.`, `%`, `-` or nothing in front of it the regex gets more complicated, and I don't know if it is worth it 😕

PR #97, opened 2016-11-21
added VersionSubscriber with creation of versions

danrot:
This PR adds a new subscriber, which creates a new version everytime a document is published. In addition to that it keeps documents checked out if necessary.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -14,7 +14,8 @@
         "jackalope/jackalope-doctrine-dbal": "~1.2.3",
         "phpunit/phpunit": "^4.5",
         "monolog/monolog": "^1.1",
-        "phpbench/phpbench": "0.10.*"
+        "phpbench/phpbench": "0.10.*",
+        "symfony/phpunit-bridge": "~2.8 || ~3.0"
     },
     "license": "MIT",
     "authors": [
--- a/lib/DocumentManager.php
+++ b/lib/DocumentManager.php
@@ -139,9 +139,11 @@ class DocumentManager implements DocumentManagerInterface
     /**
      * {@inheritdoc}
      */
-    public function restore($document, $locale, $version)
+    public function restore($document, $locale, $version, array $options = [])
     {
-        $event = new Event\RestoreEvent($document, $locale, $version);
+        $options = $this->getOptionsResolver(Events::FIND)->resolve($options);
+
+        $event = new Event\RestoreEvent($document, $locale, $version, $options);
         $this->eventDispatcher->dispatch(Events::RESTORE, $event);
     }
 
--- a/lib/DocumentManager.php
+++ b/lib/DocumentManager.php
@@ -137,6 +137,15 @@ class DocumentManager implements DocumentManagerInterface
     /**
      * {@inheritdoc}
      */
+    public function restore($document, $locale, $version)
+    {
+        $event = new Event\RestoreEvent($document, $locale, $version);
+        $this->eventDispatcher->dispatch(Events::RESTORE, $event);
+    }
+
+    /**
+     * {@inheritdoc}
+     */
     public function refresh($document)
     {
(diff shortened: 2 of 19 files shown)


composer.json
```diff
@@ -14,7 +14,8 @@
         "jackalope/jackalope-doctrine-dbal": "~1.2.3",
         "phpunit/phpunit": "^4.5",
         "monolog/monolog": "^1.1",
-        "phpbench/phpbench": "0.10.*"
+        "phpbench/phpbench": "0.10.*",
+        "symfony/phpunit-bridge": "~2.7"
```
danrot: @wachterjohannes What version should this dependency be set to now? I need it because of the `ClockMock` class.

danrot:
Tests are failing because I had to reintroduce the phpunit-bridge from Symfony for the `ClockMock` class, and some dependencies are containing deprecations again. Should we add the deprecation helper `weak` option` again?

danrot:
Nevermind, fixed it 🙈 Had just to use the correct slugifier class. However, I am still in favour of not using the Symfony CMF slugifier at all, since it doesn't add any value. If we would just use the `Ferrandini/Urlizer` wrapped with our custom interface we would keep control over these things... But that's probably another discussion and out of scope of this PR anyway.

lib/DocumentManager.php
```diff
@@ -137,6 +139,17 @@ public function removeDraft($document, $locale)
     /**
      * {@inheritdoc}
      */
+    public function restore($document, $locale, $version, array $options = [])
+    {
+        $options = $this->getOptionsResolver(Events::FIND)->resolve($options);
```
chirimoya: `Events::RESTORE`

lib/Subscriber/Behavior/VersionSubscriber.php
```diff
@@ -0,0 +1,265 @@
...
+            $versions = $nodes[$versionInformation['path']]->getPropertyValueWithDefault(static::VERSION_PROPERTY, []);
+
+            if (!array_key_exists($versionInformation['path'], $nodeVersions)) {
+                $nodeVersions[$versionInformation['path']] = $versions;
+            }
+            $nodeVersions[$versionInformation['path']][] = json_encode([
+                'locale' => $versionInformation['locale'],
+                'version' => $version->getName(),
+                'author' => $versionInformation['author'],
+                'authored' => date('c', time()),
```
chirimoya: use only `date('c')`

lib/Subscriber/Behavior/VersionSubscriber.php
```diff
@@ -0,0 +1,265 @@
...
+        if (!$this->support($event->getDocument())) {
+            $event->stopPropagation();
+
+            return;
+        }
+
+        $contentPropertyPrefix = $this->propertyEncoder->localizedContentName('', $event->getLocale());
+        $systemPropertyPrefix = $this->propertyEncoder->localizedSystemName('', $event->getLocale());
+
+        $node = $event->getNode();
```
chirimoya, line 223: add comment

lib/Subscriber/Behavior/VersionSubscriber.php
```diff
@@ -0,0 +1,265 @@
...
+        foreach ($node->getProperties() as $property) {
+            if ($this->isRestoreProperty($property->getName(), $contentPropertyPrefix, $systemPropertyPrefix)) {
+                $property->remove();
+            }
+        }
+
+        $version = $this->versionManager->getVersionHistory($node->getPath())->getVersion($event->getVersion());
+        $frozenNode = $version->getFrozenNode();
+
+        foreach ($frozenNode->getPropertiesValues() as $name => $value) {
```
chirimoya, line 236: add comment

chirimoya:
created and changed date is missing in the live session

PR #107, opened 2017-03-09
Added VersionNotFoundException

trickreich:


lib/Exception/VersionNotFoundException.php
```diff
@@ -0,0 +1,16 @@
...
+ *
+ * (c) MASSIVE ART WebServices GmbH
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace Sulu\Component\DocumentManager\Exception;
+
+class VersionNotFoundException extends DocumentManagerException
```
danrot, line 14: I prefer to not have such a small exception, but a parameterized one. So that it can be used like `throw new VersionNotFoundException($document, $version)`, and it generates the message on its own.

danrot (changes requested):

danrot (approved):

PR #108, opened 2017-03-22
Added recursive restore to allow also versions of children

wachterjohannes:
This PR extends the `VersionSubscriber` to allow also restore children.

- [x] https://github.com/jackalope/jackalope/pull/329

Requires jackalope/jackalope ^1.2.8

### DIFF
--- a/lib/Subscriber/Behavior/VersionSubscriber.php
+++ b/lib/Subscriber/Behavior/VersionSubscriber.php
@@ -285,7 +285,7 @@ class VersionSubscriber implements EventSubscriberInterface
             );
         }
 
-        // remove child-nodes which does not exists in frozen-node
+        // remove child-nodes which do not exists in frozen-node
         foreach ($node->getNodes() as $childNode) {
             if ($childNode->getDefinition()->getOnParentVersion() !== OnParentVersionAction::COPY
                 || $frozenNode->hasNode($childNode->getName())
--- a/lib/Subscriber/Behavior/VersionSubscriber.php
+++ b/lib/Subscriber/Behavior/VersionSubscriber.php
@@ -14,6 +14,7 @@ namespace Sulu\Component\DocumentManager\Subscriber\Behavior;
 use Jackalope\Version\VersionManager;
 use PHPCR\NodeInterface;
 use PHPCR\SessionInterface;
+use PHPCR\Version\OnParentVersionAction;
 use PHPCR\Version\VersionException;
 use Sulu\Component\DocumentManager\Behavior\VersionBehavior;
 use Sulu\Component\DocumentManager\Event\AbstractMappingEvent;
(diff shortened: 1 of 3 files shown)


lib/Subscriber/Behavior/VersionSubscriber.php
```diff
@@ -226,26 +229,71 @@ public function restoreProperties(RestoreEvent $event)
...
-                }
+            $this->restoreNode(
+                $node->getNode($childNode->getName()),
+                $childNode,
+                $contentPropertyPrefix,
+                $systemPropertyPrefix
+            );
+        }
+
+        // remove child-nodes which does not exists in frozen-node
```
danrot: `which do not exist in frozen node`

tests/Unit/Subscriber/Behavior/VersionSubscriberTest.php
```diff
@@ -343,6 +345,7 @@ public function testRestoreLocalizedProperties()
         $property3 = $this->prophesize(PropertyInterface::class);
         $property3->getName()->willReturn('jcr:uuid');
         $node->getProperties()->willReturn([$property1->reveal(), $property2->reveal(), $property3->reveal()]);
+        $node->getNodes()->willReturn([]);
```
danrot, line 348: Please add also a test where the node has children, which should not be copied (as it is the case with the pages).

danrot (changes requested):

wachterjohannes:
@danrot fixed

tests/Unit/Subscriber/Behavior/VersionSubscriberTest.php
```diff
@@ -457,4 +457,48 @@ function () use ($node, $newChild1Node, $newChild2Node, $newChild3Node) {
...
+        $event->getDocument()->willReturn($document->reveal());
+        $event->getNode()->willReturn($node->reveal());
+        $event->getVersion()->willReturn('1.0');
+        $event->getLocale()->willReturn('de');
+
+        $this->versionManager->getVersionHistory('/node')->willReturn($versionHistory->reveal());
+        $versionHistory->getVersion('1.0')->willReturn($version->reveal());
+        $version->getFrozenNode()->willReturn($frozenNode->reveal());
+
+        $this->versionSubscriber->restoreProperties($event->reveal());
```
danrot, line 506: There's not a single assertion in this test, so we cannot be sure that the child node has not been touched...

danrot (changes requested):

tests/Unit/Subscriber/Behavior/VersionSubscriberTest.php
wachterjohannes: it's not explicit but we could. every not mocked method call would let this test fail - but i will add some assertions.

tests/Unit/Subscriber/Behavior/VersionSubscriberTest.php
```diff
@@ -457,4 +457,52 @@ function () use ($node, $newChild1Node, $newChild2Node, $newChild3Node) {
...
+        $event->getDocument()->willReturn($document->reveal());
+        $event->getNode()->willReturn($node->reveal());
+        $event->getVersion()->willReturn('1.0');
+        $event->getLocale()->willReturn('de');
+
+        $this->versionManager->getVersionHistory('/node')->willReturn($versionHistory->reveal());
+        $versionHistory->getVersion('1.0')->willReturn($version->reveal());
+        $version->getFrozenNode()->willReturn($frozenNode->reveal());
+
+        // the child-node should not be touched
```
wachterjohannes, line 502: @danrot i have added assertions

danrot (approved):

PR #109, opened 2017-04-11
Added metadata to configure remove-live

wachterjohannes:
This PR introduces a boolean flag which will be used by sulu to configure the behavior for removing instances of this document.

See https://github.com/sulu/sulu/pull/3302

### DIFF
--- a/lib/Metadata.php
+++ b/lib/Metadata.php
@@ -44,6 +44,11 @@ class Metadata
     private $fieldMappings = [];
 
     /**
+     * @var bool
+     */
+    private $syncRemoveLive = true;
+
+    /**
      * Add a field mapping for field with given name, for example.
      *
      * ````
@@ -87,6 +92,8 @@ class Metadata
     }
 
     /**
+     * Returns class.
+     *
      * @return string
      */
     public function getClass()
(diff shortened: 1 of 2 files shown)


lib/Metadata.php
```diff
@@ -44,6 +44,11 @@ class Metadata
     private $fieldMappings = [];
 
     /**
+     * @var bool
+     */
+    private $removeLive = true;
```
danrot: I am somehow missing the connection that this is done automatically when the draft is deleted. Maybe `syncRemoveAcrossWorkspaces`? Although this is quite a long one... Or `syncRemoveLive`?

danrot: Not completely sure about this now, but `removeLive` could also be interpreted as never having a live node at all.

danrot (changes requested):

lib/Metadata.php
wachterjohannes: @danrot in my opinion, the `syncRemovLive` sounds good but not perfect ... but i cannot think of any other name. what should i do? (:

danrot: Let's take `syncRemoveLive`, since we don't know a better name...

wachterjohannes: @danrot fixed

wachterjohannes:
@danrot fixed

danrot (approved):

PR #113, opened 2017-05-22
updated proxy manager to be compatible with PHP7

danrot:
Made an update of `ocramius/proxy-manager` to the latest version, in order to be compatible with PHP7.

Merge should wait until it has been tested in a complete setup.

### DIFF
--- a/tests/Unit/ProxyFactoryTest.php
+++ b/tests/Unit/ProxyFactoryTest.php
@@ -28,6 +28,46 @@ use Symfony\Component\EventDispatcher\EventDispatcherInterface;
 class ProxyFactoryTest extends \PHPUnit_Framework_TestCase
 {
     /**
+     * @var NodeInterface
+     */
+    private $node;
+
+    /**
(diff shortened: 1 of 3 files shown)


tests/Unit/ProxyFactoryTest.php
```diff
@@ -152,8 +192,10 @@ public function setParent($parent)
 
 class TestProxyDocumentProxy
 {
+    private $title = 'title';
+
     public function getTitle()
     {
-        return 'Hello';
+        return $this->title;
```
danrot, line 199: Methods are not proxied anymore, instead only properties have a proxy. Therefore it is necessary to use a property here as well.

danrot:
The new proxy-manager is also using an evaluation strategy as the new default. That means that the proxies are actually not written to a file, which is not advisable to do so in production mode, but in the development mode it has the advantage that we cache less.

I think I'll just change it to another strategy writing to disk instead for now.

PR #114, opened 2017-05-22
Transliterating slugifier to support non-latin characters

yukoff:
This is done in favor of sulu/sulu#3367
* use [behat/transliterator](https://packagist.org/packages/behat/transliterator) instead of [aferrandini/urlizer](https://packagist.org/packages/aferrandini/urlizer) - they are very similar and have same api (or at least very similar, anyway both have `urlize()`/`transliterate()`), but `behat/transliterator` is updated to a more recent version of [Text::Unidecode](http://search.cpan.org/~sburke/Text-Unidecode-1.27/lib/Text/Unidecode.pm) data, which both are based upon
* extend unit test to unhide the undergoing problem

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@ CHANGELOG for Sulu Document Manager
 ===================================
 
 * dev-develop
+    * BUGFIX      #113 Updated ProxyManager to be compatible with PHP 7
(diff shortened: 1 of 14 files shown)


yukoff:
Sorry, should be against develop I guess

PR #115, opened 2017-05-22
Transliterating slugifier to support non-latin characters

yukoff:
This is done in favor of sulu/sulu#3367
* use [behat/transliterator](https://packagist.org/packages/behat/transliterator) instead of [aferrandini/urlizer](https://packagist.org/packages/aferrandini/urlizer) - they are very similar and have same api (or at least very similar, anyway both have `urlize()`/`transliterate()`), but `behat/transliterator` is updated to a more recent version of [Text::Unidecode](http://search.cpan.org/~sburke/Text-Unidecode-1.27/lib/Text/Unidecode.pm) data, which both are based upon
* extend unit test to unhide the undergoing problem
* add basic tests for extended latin and non-latin symbols (ukrainian cyrillic, japanese, chinese (simplified) and russian cyrillic)

### DIFF
--- a/symfony-di/core.xml
+++ b/symfony-di/core.xml
@@ -43,7 +43,7 @@
         </service>
 
         <service id="sulu_document_manager.slugifier" class="Symfony\Cmf\Api\Slugifier\CallbackSlugifier" public="false">
-            <argument>Ferrandini\Urlizer::urlize</argument>
+            <argument>Behat\Transliterator\Transliterator::transliterate</argument>
         </service>
 
         <service id="sulu_document_manager.node_name_slugifier"
--- a/composer.json
+++ b/composer.json
@@ -8,7 +8,7 @@
         "symfony/options-resolver": "~2.8 || ~3.0",
         "symfony-cmf/slugifier-api": "~1.0",
         "ocramius/proxy-manager": "~1.0 | ~2.0",
-        "aferrandini/urlizer": "~1.0.0"
+        "behat/transliterator": "^1.2.0"
     },
     "require-dev": {
         "jackalope/jackalope-doctrine-dbal": "~1.2.3",
--- a/tests/Unit/Slugifier/NodeNameSlugifierTest.php
+++ b/tests/Unit/Slugifier/NodeNameSlugifierTest.php
@@ -12,10 +12,17 @@
 namespace Sulu\Component\DocumentManager\tests\Unit\Slugifier;
 
 use Sulu\Component\DocumentManager\Slugifier\NodeNameSlugifier;
+use Sulu\Component\DocumentManager\Tests\Bootstrap;
 use Symfony\Cmf\Api\Slugifier\SlugifierInterface;
+use Symfony\Component\DependencyInjection\Container;
 
 class NodeNameSlugifierTest extends \PHPUnit_Framework_TestCase
 {
+    /**
+     * @var Container
+     */
+    private $container;
+
     /**
      * @var SlugifierInterface
      */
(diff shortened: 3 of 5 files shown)


symfony-di/core.xml
```diff
@@ -43,7 +43,7 @@
         </service>
 
         <service id="sulu_document_manager.slugifier" class="Symfony\Cmf\Api\Slugifier\CallbackSlugifier" public="false">
-            <argument>Ferrandini\Urlizer::urlize</argument>
+            <argument>Behat\Transliterator\Transliterator::transliterate</argument>
```
yukoff, line 46: This works, but as for me a way better solution would be to implement something like IntlLocaleAwareSlugifier, wich would utilize `ext-intl` and it's `Transliterator::transliterate()`

wachterjohannes:
@yukoff code looks good to me. it also should fix the resource-locator bug with arabian titles - which we had a few months ago! but @danrot what do you thing about an additional dependency? should we try to get rid of urlizer? i think we use it at some other places!

yukoff:
Actually `behat/transliterator` is used by [`gedmo/doctrine-extensions`](https://packagist.org/packages/gedmo/doctrine-extensions) (used by sulu) and anyway installed as a dependency, so shouldn't be a problem.

yukoff:
I'm preparing corresponding changes for sulu itself - there are some places where it is (urlizer) being used. From what I've found so far Urlizer is used by SuluCustomUrlBundle functional controller test (directly, by means of `use ...;`) and by `Sulu\Bundle\RouteBundle\Generator\RouteGenerator` (via `sulu_document_manager.slugifier` service).

danrot:
The other issue is https://github.com/sulu/sulu-standard/issues/698. Well, looks like it actually solves the problem, so I would risk changing the dependency.

However, does this `transliterate` method also urlize the string? Haven't digged into it to deeply, but it sounds like it more about just mapping special characters to the ascii equivalent.

yukoff:
To outline difference I've prepared a simple test for arabic string for `test article` from [google translate](https://translate.google.com.ua/#en/ar/test%20article) ;)

To understand results this could be helpful:
```
use Ferrandini\Urlizer;
use Behat\Transliterator\Transliterator as BehatTransliterator;
use Transliterator as IntlTransliterator;
$tl = IntlTransliterator::create('Any-Latin; Latin-ASCII; NFD; [:Nonspacing Mark:] Remove; NFKC;');
```

And here are results:
```
### $tl->transliterate: akhtbar almadt
### $tl->transliterate + BehatTransliterator::urlize: akhtbar-almadt
### Urlizer::urlize:
### Urlizer::transliterate: ?????? ??????
### BehatTransliterator::urlize:
### BehatTransliterator::transliterate: khtbr-lmd
```

wachterjohannes:
@yukoff @danrot this looks good (: it is not an additional dependency so i would go with it.

danrot:
What I am worrying about is that `BehatTransliterator::transliterate` is not always returning a string being valid for usage in a URL.... I am a bit skeptical because there is also a [`BehatTransliterator::urlize`](https://github.com/Behat/Transliterator/blob/54f7871/src/Behat/Transliterator/Transliterator.php#L414) method, wich does not work correctly with the eastern languages according to the comment.

yukoff:
@danrot `transliterate()` and `urlize()` behaves similar (even have same function signature and default param value), except `urlize()` doesn't transliterate string and for anything non-latin will return empty string :(

From the code:
```php
    public static function urlize($text, $separator = '-')
    {
        $text = self::unaccent($text);
        return self::postProcessText($text, $separator);
    }

    public static function transliterate($text, $separator = '-')
    {
        if (preg_match('/[\x80-\xff]/', $text) && self::validUtf8($text)) {
            $text = self::utf8ToAscii($text);
        }
        return self::postProcessText($text, $separator);
    }
```

yukoff:
@danrot Yeah, I understand your skepticizm - I'm feeling the same here, and actually that's why I mentioned intl's [Transliterator](https://secure.php.net/manual/en/class.transliterator.php), which might be the best bet. But that wouldn't be a straight drop-in replacement (passing a callable to CallbackSlugifier), some additional work is required and to provide most accurate transliteration it should be locale-aware (I guess).

danrot:
Actually I am a little bit confused by the naming of the `Transliterator`. It's a little bit strange that `transliterate` also seems to replace ` ` with `-`. I would say if you have time to implement the other parts in `sulu/sulu`, and we'll test it heavily to make sure everything works.

yukoff:
Yeah, naming is quite confusing - have to dig into code to get the answers :) I prepared a WIP PR for sulu to review undergoing changes while I'm finishing with ResourceLocator part - sulu/sulu#3372

